PR #4, opened 2016-10-03
Scan for min max sample

kjellkongsvik:
Example program

### DIFF
--- a/examples/scan_min_max.py
+++ b/examples/scan_min_max.py
@@ -1,0 +1,24 @@
+import sys
+import segyio
+import numpy as np
+
+
+def main():
+    if len(sys.argv) < 2:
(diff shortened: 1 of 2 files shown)


examples/scan_min_max.py
```diff
@@ -0,0 +1,24 @@
...
+    min_value = sys.float_info.max
+    max_value = sys.float_info.min
+
+    with segyio.open(segyfile) as f:
+        for trace in f.trace:
+            min_value = min(np.amin(trace), min_value)
+            max_value = max(np.amin(trace), max_value)
+
+    print "min: {}".format(min_value)
+    print "max: {}".format(max_value)
```
pgdr: Consider adding `from __future__ import print_function` and use Python 3 compatible print.  (In general, it's good to try to be Python 3 compatible.)

PR #29, opened 2016-10-19
Portability, Windows support, Windows CI, fewer warnings and cmake cleanup

jokva:
Ok, so quite a lot is going on in this PR, but here's the short-and-sweet synopsis:

* cmake files are slightly restructured to be easier to follow - configurations are largely more local, i.e. no python details leak outside of the python cmake file.
* The C applications are moved as a consequence.
* Debug builds now compile with quite a few warnings. I expect these warnings to be tuned slightly as we move forward, but it's a good starting point I think.
* Support for MSVC, and by extension windows, has been added.
* The C layer has received quite a few portability improvements (mostly demonstrated by crashes on windows)
* Windows CI (appveyor) support.
* Plenty of warning fixes, including a few cases that would alter the program enough to crash on windows.
* Some byte-level functionality is now handled by libraries rather than by hand.

I of course welcome any feedback and criticism. I can, if needed, split this PR up, however I don't really see the point of that.


applications/CMakeLists.txt
```diff
@@ -3,3 +3,14 @@ configure_file(segyviewer.py segyviewer COPYONLY)
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/segyviewer
```
jepebe: Should only be installed if Python is built...

src/segyio/segy.h
```diff
@@ -54,7 +54,7 @@ int segy_sample_indexes(FILE* fp, double* buf, double t0, size_t count);
 
 /* text header operations */
 int segy_read_textheader(FILE *, char *buf);
-unsigned int segy_textheader_size();
+int segy_textheader_size();
```
jepebe: This means that the return type in _segyio.c should be updated.

PR #34, opened 2016-10-20
Added __version__ to __init__

pgdr:


pgdr:
so that I can do
```
In [1]: import segyio

In [2]: segyio.__version__
Out[2]: '1.0.2'
```

and thus use
```
python_module(segyio 1.0.2)
if (NOT DEFINED PY_segyio)
  message(SEND_ERROR "segyio module not found - Python wrappers not enabled")
endif()
```
from [Statoil/pycmake](https://github.com/Statoil/pycmake)

jepebe:
I think `setup.py` needs some love here. I _manually_ enter a version into `setup.py`. Maybe version number should be in a `VERSION` file of some sort that both. `CMAKE` and `setup.py` could fetch the version from?

PR #38, opened 2016-10-27
Trace.raw mode for eager reading of traces

jokva:
Trace.raw mode for eager reading of traces

Adds support for the raw trace sub mode that does eager reading of
traces into an MxN numpy array - a useful feature for segyview and other
visualisation tools when it is known beforehand that the file will fit
in memory.

### DIFF
--- a/python/segyio/_raw_trace.py
+++ b/python/segyio/_raw_trace.py
@@ -1,0 +1,19 @@
+import numpy as np
+import segyio
+
+class RawTrace(object):
+    def __init__(self, trace):
+        self.trace = trace
+
+    def __getitem__(self, index):
+        """ :rtype: numpy.ndarray """
+        buf = None
+        if isinstance(index, slice):
+            f = self.trace._file
+            start, stop, step = index.indices(f.tracecount)
+            mstart, mstop = min(start, stop), max(start, stop)
+            length = max(0, (mstop - mstart + (step - (1 if step > 0 else -1))))
+            buf = np.zeros(shape = (length, f.samples), dtype = np.single)
+
+        return self.trace._readtr(index, buf)
+
--- a/python/segyio/_segyio.c
+++ b/python/segyio/_segyio.c
@@ -705,18 +705,29 @@ static PyObject *py_fread_trace0(PyObject *self, PyObject *args) {
 static PyObject *py_read_trace(PyObject *self, PyObject *args) {
     errno = 0;
     PyObject *file_capsule = NULL;
-    unsigned int trace_no;
+    PyObject *trace_no;
     PyObject *buffer_out;
+    int trace_count;
     long trace0;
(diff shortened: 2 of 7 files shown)


python/segyio/_raw_trace.py
```diff
@@ -0,0 +1,19 @@
...
+        """ :rtype: numpy.ndarray """
+        buf = None
+        if isinstance(index, slice):
+            f = self.trace._file
+            start, stop, step = index.indices(f.tracecount)
+            mstart, mstop = min(start, stop), max(start, stop)
+            length = max(0, (mstop - mstart + (step - (1 if step > 0 else -1))))
+            buf = np.zeros(shape = (length, f.samples), dtype = np.single)
+
+        return self.trace._readtr(index, buf)
```
jepebe: What if index is int then but is None. How is this handled?

python/segyio/_segyio.c
```diff
@@ -705,33 +705,62 @@ static PyObject *py_fread_trace0(PyObject *self, PyObject *args) {
 static PyObject *py_read_trace(PyObject *self, PyObject *args) {
     errno = 0;
     PyObject *file_capsule = NULL;
-    unsigned int trace_no;
+    PyObject *trace_no;
     PyObject *buffer_out;
+    Py_ssize_t start, stop, step, length;
```
jepebe: Move!!!

python/segyio/segy.py
```diff
@@ -302,6 +302,23 @@ def trace(self):
...
+            (especially memory-intensive) operations then this feature is
+            unlikely to help.
+
+            Read the first 10 traces::
+                >>> f.trace.raw[0:10]
+
+            Read *all* traces to memory::
+                >>> f.trace.raw[:]
+
+            Read every other trace traces to memory::
```
jepebe: trace traces?

python/segyio/segy.py
```diff
@@ -302,6 +302,23 @@ def trace(self):
             Fill the file with one trace (filled with zeros)::
                 >>> tr = np.zeros(f.samples)
                 >>> f.trace = itertools.repeat(tr)
+
+            For advanced users: sometimes you want to load the entire segy file
+            to memory and apply your own structural manipulations or operations
+            on it. Some segy files are very large and may not fit, in which
+            case this feature will break down. This is an optimisation feature.
+            If you're doing something trace-by-trace and performing non-trivial
```
jepebe: I think the non-trivial memory intensive bit does not make sense. But maybe if you work trace-by-trace or line-by-line...

python/segyio/_raw_trace.py
jokva: By readtr which again calls _segyio.read_trace which soon also checks for None (both for slice and int).

jokva:
Adress'd.

PR #43, opened 2016-10-31
Trying to move the python code to be py3k compatible

asbjorn:
More Python 3.X changes to the python code

### DIFF
--- a/python/segyio/_header.py
+++ b/python/segyio/_header.py
@@ -23,7 +23,7 @@ class Header(object):
             gen_buf = self._header_buffer(buf)
 
             def gen():
-                for i in xrange(*traceno.indices(self.segy.tracecount)):
+                for i in range(*traceno.indices(self.segy.tracecount)):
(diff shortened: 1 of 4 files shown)


pgdr:
Then we should add `from builtins import range` or something similar to maintain `xrange`-behavior.

jokva:
Something like that. The range function itself isn't so bad (the lists are never *that* big and they don't strictly require generator behaviour - the `izip`s do.

jokva:
That being said, thanks for this effort! Are you trying to run it with python3? What other problems have you faced?

asbjorn:
@pgdr: Yes, to get the `range` backported to Python 2.X the `builtins.range` can be really handy.

You will need to update the requirements.txt file for that though - but, still :+1: 

Will that work for both Python 2.X or Python 3.X or do you need to do a check like this:
```import sys
if sys.version_info < (3.0):
    from builtins import range
```

PR #56, opened 2016-11-10
make-ps-file.py; example program for pre-stack

jokva:
The make-ps-file.py program will create a prestack file using more or
less the same scheme as make-file.py. Also adds one outputted prestack
file for testing purposes.

Relies on https://github.com/Statoil/SegyIO/pull/54


jokva:
Made obsolete by https://github.com/Statoil/SegyIO/pull/58

PR #63, opened 2016-11-18
Read dt in python. Closes #62

### LINKED ISSUE #62: Access dt from python

kjellkongsvik:


python/segyio/segy.py
```diff
@@ -28,7 +28,7 @@
 
 class SegyFile(object):
 
-    def __init__(self, filename, mode, iline=189, xline=193, t0=1111.0):
+    def __init__(self, filename, mode, iline=189, xline=193):
```
jepebe: I think that it should be possible to set `t0` unless it is trivial and guaranteed to get that value from the headers. The default should probably be `none` to allow to find in headers and default to `0`, while a value of `0` is specific.

python/segyio/segy.py
```diff
@@ -46,6 +46,7 @@ def __init__(self, filename, mode, iline=189, xline=193, t0=1111.0):
         self._offsets = None
         self._ext_headers = None
         self._samples = None
+        self._dt = None
```
jepebe: Should it be possible to override `dt`?

jokva: Not with our current needs.

python/segyio/segy.py
jokva: Seems reasonable to me. I don't think it's a blocker though.

python/segyio/_segyio.c
```diff
@@ -479,6 +479,25 @@ static PyObject *py_trace_bsize(PyObject *self, PyObject *args) {
...
+    PyArg_ParseTuple(args, "Od", &file_capsule, &dt);
+    segy_file *p_FILE = get_FILE_pointer_from_capsule(file_capsule);
+
+    if (PyErr_Occurred()) { return NULL; }
+
+    int error = segy_sample_interval(p_FILE, &dt);
+    if (error != 0) { return py_handle_segy_error(error, errno); }
+
+    PyObject *dict = PyDict_New();
+    PyDict_SetItemString(dict, "dt", Py_BuildValue("d", dt));
```
jokva: Why does this return a dict over a plain int?

src/segyio/segy.c
```diff
@@ -658,39 +658,36 @@ static int segy_sample_interval( segy_file* fp, double* dt) {
...
-    } else if (trace_header_dt == 0.0) {
-        fprintf(stderr, "Trace sampling rate in trace header set to 0.0. Will use SEGY header sampling rate of: %f\n", binary_header_dt);
-        *dt = binary_header_dt;
-    } else if (trace_header_dt != binary_header_dt) {
-        fprintf(stderr, "Trace sampling rate in SEGY header and trace header are not equal. Will use SEGY header sampling rate of: %f\n",
-                binary_header_dt);
-        *dt = binary_header_dt;
+    // microseconds: us
+    int _binary_header_dt_us;
+    int _trace_header_dt_us;
```
jokva: These symbols (underscore prefix'd) are reserved for the language implementation.

src/segyio/segy.c
```diff
@@ -658,39 +658,36 @@ static int segy_sample_interval( segy_file* fp, double* dt) {
...
+
+    segy_get_bfield(bin_header, BIN_Interval, &_binary_header_dt_us);
+    segy_get_field(trace_header, TRACE_SAMPLE_INTERVAL, &_trace_header_dt_us);
+
+    // milliseconds: ms
+    double binary_header_dt_ms = _binary_header_dt_us/1000.0;
+    double trace_header_dt_ms = _trace_header_dt_us/1000.0;
+
+    if (_binary_header_dt_us == 0 && _trace_header_dt_us == 0) {
+        fprintf(stderr, "Trace sampling rate in SEGY header and first trace header set to 0.0. Will not set.\n");
```
jokva: We shouldn't fprintf from inside the library. It's up to the wrapper code to report errors in a fashion that suits them

src/segyio/segy.c
```diff
@@ -658,39 +658,36 @@ static int segy_sample_interval( segy_file* fp, double* dt) {
...
+    } else {
+        *dt = trace_header_dt_ms;
     }
 
     return 0;
 
 }
 
-int segy_sample_indexes( segy_file* fp, double* buf, double t0, size_t count) {
+int segy_sample_indexes( segy_file* fp, double* buf, double t0, double dt, size_t count) {
```
jokva: Does this have to be `double`? The rest of the library assumes float.

kjellkongsvik: agreed, but was somewhat mixed usage of double and float. t0 and sample_indexes is double. Can of course change all that.

src/segyio/segy.c
kjellkongsvik: Agreed. Will silently use fallback dt

python/segyio/_segyio.c
jokva: What error objects? You're not returning one now. You can also use exceptions.

src/segyio/segy.c
jokva: Not necessarily silently (you can introduce a new error code, for instance).

python/segyio/_segyio.c
kjellkongsvik: if (error != 0) { return py_handle_segy_error(error, errno); }

jokva: That will set an exception, not return an object. That function returns NULL.

src/segyio/segy.c
kjellkongsvik: That is possible. But probably not interesting.

jepebe:
Looks good! Can a segy file, technically, have different *dt* for every trace?

kjellkongsvik:
@jepebe yes, can have a different dt for every trace. We only check binary header and first trace.

PR #71, opened 2017-01-11
Align enumerations to established names

jokva:
Change names to match those in existing code bases so that the Segy
prefixed functions are drop-in compatible.


jokva:
Made obsolete by https://github.com/Statoil/SegyIO/pull/72

PR #77, opened 2017-02-09
Reorganise makefiles and CI

jokva:
Synopsis:
* New directory layout.
* Makefiles rewritten, export and install added
* Tests moved to language directory
* Windows DLLs expose symbols (with .def file)
* cmake module paths no longer absolute
* CI (appveyor.yml and .travis.yml) are more language agnostic
* segyio-config.cmake created
* Embeddable in other projects

The top level directory now contains language-specific sub directories
and global files such as changelog, licence, and readme.

Tests and example programs are moved to their language directory, rather
than being bundled together in tests/. This is largely to help keeping
makefiles very language or sub-project specific, so minimal information
need to be shared between sibling directories.

Test data is still global, and is moved to /test-data. Every language's
makefile is responsible for copying the test files it needs to the place
it needs them.

The python.cmake file has been heavily updated and generalised, and
makes fewer assumptions about directory layout. It is also aware of
potential C/C++ targets and handles all python-specific things, meaning
the makefile for the python support can be very segyio oriented.

The segyspec C functions are now considered an internal matlab detail
and are moved into the mex directory.

The C core layout has changed slightly. It is organised under lib/, and
public headers are in the include/segyio directory. This separation is
useful, since the segyio-* targets export their public headers
(include/segyio), and not its private headers. This allows easier
inclusion of segyio in other cmake based projects, and creates a clear
separation of public and private interface.

Ultimately, this means that projects can depend on segyio by either
using segyio-config.cmake or source-wise by doing add_subdirectory. This
was tested by writing a tiny program that depended on segyio with
add_subdirectory(segyio) in the makefile, where segyio was a fresh
checkout of this commit.

Finally, The CI files no longer assume python is the only language to
ever be supported.


jokva:
The diff if very inflated because of some files changing directory. The major changes are in the CMakeLists.txt files.

cmake/python.cmake
```diff
@@ -1,50 +1,152 @@
...
+#
+# Create a new target <target>, analogous to add_library. Creates a python
+# package <name>, optionally at the path specified with PATH. If SUBDIR <dir>
+# is used, then the files will be copied to $root/dir/*, in order to create
+# python sub namespaces - if subdir is not used then all files will be put in
+# $root/*. SOURCES <src> is the python files to be copied to the directory in
+# question, and <tgt> are regular cmake libraries (targets created by
+# add_library).
+#
+# Wuen the APPEND option is used, the files and targets given will be added
```
kjellkongsvik: typo: -> when

PR #81, opened 2017-02-13
Build+link respecting BUILD_SHARED_LIBS

jokva:
Both targets will always be built, but the target segyio will alias the
shared or static library depending on BUILD_SHARED_LIBS being on or off.


jokva:
This isn't ready yet - I need to resolve how matlab (which expects to link shared) can deal with a possibly position-dependent library.

jepebe:
Please expose the `BUILD_SHARED_LIBS` flag so that it shows up with `ccmake`. For example: `option(BUILD_SHARED_LIBS "Build libs shared or static" ON)`

jokva:
Rebased and updated.

PR #83, opened 2017-02-14
Handle 4G+ files on sizeof(long) == 4 platforms.

jokva:
The use of size_t for file positions uses an implicit conversion to
long for calls to fseek and friends. Long is *not* guaranteed to be
64-bit, so this effectively restricts portable support for very large
files.

segy_seek has been rewritten to consider LONG_MAX and doing repeated
relative seeks should the position overflow a long. On most platform
this path should not be taken and optimised out.

### DIFF
--- a/cmake/check_includes.cmake
+++ b/cmake/check_includes.cmake
@@ -1,4 +1,5 @@
 include(CheckIncludeFile)
+include(CheckFunctionExists)
 
 # Portability checks; look for htons function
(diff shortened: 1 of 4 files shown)


jokva:
I've tested the *negative* here (where ftell returns a long) and I can confirm that it fails as expected. This gives new confidence to the 4G test which I think was subtly broken on systems where `sizeof(long) == 4`

PR #85, opened 2017-02-15
Remove unsigned in interface

jokva:
The mixture of signed and unsigned and various int types was largely an accident, and this PR attempts to clean up most of this. In short:

* All signatures have been reviewed and unless there's a good reason for some variable *not* to be int, it's int.
* The trace0 function still uses long, in order to be immediately associated with `fseek`. Overflow here isn't a problem since the trace0 is rather early in the file.
* Some explicit casts are added to silence conversion warnings that are deemed safe.
* `sample_indexes` have been renamed to `sample_indices` to be consistent with the rest of segyio, and has its argument shuffled around. Additionally, the arguments for this functions are now floats (as that's what we're working with in segyio) instead of doubles.

### DIFF
--- a/applications/segyinfo.c
+++ b/applications/segyinfo.c
@@ -51,7 +51,7 @@ int main(int argc, char* argv[]) {
     const int format = segy_format( header );
     const int samples = segy_samples( header );
     const long trace0 = segy_trace0( header );
-    const unsigned int trace_bsize = segy_trace_bsize( samples );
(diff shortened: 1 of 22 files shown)


jokva:
Just noticed that there's also a cheeky matlab-cmake commit in this set. It simply moves the MATLAB_ROOT/bin path higher in the search list so it's easier to inject a custom, location-specific matlab root.

PR #90, opened 2017-02-22
Subtrace support for start/stop/step

jokva:
Replaces the first-last interface of read/write_subtr with
start/stop/step, analoguous to python slices. With a step of 1, the most
common case, this covers the old chunk read, but the step allows greater
flexibility and maps nicely to features commonly found in higher-level
languages.


jokva:
@kwko Test uploaded.

kwko:
It appears that I have been accidentally tagged?

jokva:
@kwko Seems so indeed, 'twas intended for @kjellkongsvik. Fault of aggressive autocomplete, my bad!

PR #91, opened 2017-02-22
Correct first/last index in mex/put_traces

jokva:
The old computation was straight-up wrong - now uses the same algorithm
as get_traces.


kjellkongsvik:
Looks good, but should there not be a test for this error?

PR #92, opened 2017-02-23
Avoid overflow on large chunks in segy_to_native

jokva:
When large amounts of traces are collected in one go, it usually means a
high number of samples too. This risks integer overflow, typically
happening for files >8G, which are common. segy_to/from_native taking
long long for size fixes this issue.


jokva:
Rebase'd.

jokva:
I'll merge once the checks are through.

PR #93, opened 2017-02-28
tools.collect and tools.cube

jokva:
Collect handles the common case of copying a series of traces, lines,
slices into one stacked numpy array.

Cube attempts to read a full cube from a file or file handle and returns
the data as a 3D numpy array.


jokva:
Merging once green.

PR #100, opened 2017-03-02
Plug buffer leaks.

jokva:
Under some error conditions certain buffers would not be properly
free'd.


python/segyio/_segyio.c
```diff
@@ -761,14 +761,14 @@ static Py_buffer check_and_get_buffer(PyObject *object, const char *name, unsign
     if (strcmp(buffer.format, "i") != 0) {
         PyErr_Format(PyExc_TypeError, "The destination for %s is not a buffer object of type 'intc'", name);
         PyBuffer_Release(&buffer);
-        return buffer;
```
kjellkongsvik: also return zero_buffer above for clarity

jokva:
Tests green and zero-buffer returned as requested. Lacks only approval.

PR #102, opened 2017-03-08
file.samples is list; samples are not always time

jokva:
Analoguous to file.xlines, file.ilines, file.samples is now a list of
samples and their identifiers. If possible, this is a list derived from
the binary- and trace headers with a recording delay t0 and a sample
interval. If these can't be figured out they're assumed to be 0 and
4000μs, 4ms, respectively.

As with the other identifiers, the cardinality is now obtained with
len().

The segy_sample_interval function would always assume that the interval
between sample describes distance-in-time (ms), but this is means a loss
of information when the data was in meters (which would be interpreted
as millimeters). Instead, this function now returns the inferred
interval without modifications, leaving it up to the caller to determine
what the interval describes, and typically has more information
available.

This means that the sample_indexes function in python.tools now changes
behaviour and uses microseconds (as per the specification) rather than
milliseconds. To support the typical use of this function,
segyio.tools.scale_samples has been added. This function uses heuristics
to determine if the sample interval is in distance (meters) or time
(milliseconds), scales the samples accordingly and returns a tuple of
samples and their units.

### DIFF
--- a/python/segyio/segy.py
+++ b/python/segyio/segy.py
@@ -477,12 +477,16 @@ class SegyFile(object):
         return buf
 
     def _line_buffer(self, length, buf=None):
-        shape = (length, self.samples)
+        shape = (length, len(self.samples))
         return self._shape_buffer(shape, buf)
 
     def _fread_line(self, trace0, length, stride, buf):
         offsets = len(self.offsets)
-        return _segyio.read_line(self.xfd, trace0, length, stride, offsets, buf, self._tr0, self._bsz, self._fmt, self.samples)
+        return _segyio.read_line(self.xfd, trace0,
+                                length, stride, offsets,
+                                buf,
+                                self._tr0, self._bsz,
+                                self._fmt, len(self.samples))
 
     @property
     def ilines(self):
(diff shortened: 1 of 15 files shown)


python/segyio/segy.py
```diff
@@ -832,16 +836,16 @@ def depth_slice(self):
...
+        other_indices = np.asarray([0], dtype=np.intc)
         buffn = self._depth_buffer
 
         slice_trace_count = self._iline_length * self._xline_length
         offsets = len(self.offsets)
         tr0 = self._tr0
         bsz = self._bsz
         fmt = self._fmt
-        samples = self.samples
+        samples = len(self.samples)
```
kjellkongsvik: rename samples, confusing. -> e.g. sample_count

python/segyio/tools.py
```diff
@@ -117,6 +117,37 @@ def cube(f):
     ilsort = f.sorting == segyio.TraceSortingFormat.INLINE_SORTING
     fast = f.ilines if ilsort else f.xlines
     slow = f.xlines if ilsort else f.ilines
-    fast, slow, offs, samples = len(fast), len(slow), len(f.offsets), f.samples
+    fast, slow, offs = len(fast), len(slow), len(f.offsets)
+    samples = len(f.samples)
```
kjellkongsvik: rename samples, confusing. -> e.g. sample_count

python/test/segy.py
```diff
@@ -96,31 +96,33 @@ def test_xline_22(self):
 
             data = f.xline[22]
 
+            samples = len(f.samples)
```
kjellkongsvik: rename samples, confusing. -> e.g. sample_count

jokva:
Alright, that should do it.

jokva:
Green please.

PR #103, opened 2017-03-09
Fix off-by-one error in non-mmap field_forall

jokva:
Fixes an off-by-one error that wasn't caught by tests because our test
files are mostly zero and the mis-aligned read happened to not affect
the result.

The issue stems from get_field being aware of the 'field' argument being
1-index'd, but the preceeding fread and its buffer were not. seek+read
is simply made aware of the 1-indexing.

Our test files headers are mostly zeros. The field-forall function would
read from byte 189, but should've started at byte 188, and the test data
happens to have all bits in the least-significant byte. When this is
copied to the local buffer, it would then be *written* with the same
mis-aligned offset, and since this didn't interact with any other bytes
(they're all set to zero anyway), the read did not pick up on it. This
is verifyable by reading the bytes into an int and using ntohl() or by
inspecting the memory with a debugger.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -549,15 +549,18 @@ int segy_field_forall( segy_file* fp,
     /*
      * non-mmap path. Doing multiple freads is slow, so instead the *actual*
      * offset is computed, not just the start of the header, and that's copied
-     * into the correct offset in our local buffer.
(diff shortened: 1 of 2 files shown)


jokva:
Green please.

PR #114, opened 2017-03-24
Make trace.__getitem__ accept more int-like types

jokva:
Under some conditions, perfectly fine int-like types would raise a type
error when passed to getitem. Instead of checking for isinstance(int),
a slice or an int-like type is expected, and we internally handle the
conversion to an actual int.

### DIFF
--- a/python/test/segy.py
+++ b/python/test/segy.py
@@ -572,6 +572,17 @@ class TestSegy(TestCase):
             self.assertListEqual(list(f.attributes(189)[:]),
                                  [(i // 5) + 1 for i in range(len(f.trace))])
 
+    def test_traceaccess_from_array(self):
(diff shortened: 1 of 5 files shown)


python/test/segy.py
```diff
@@ -572,6 +572,12 @@ def test_open_fails_unstructured(self):
             self.assertListEqual(list(f.attributes(189)[:]),
                                  [(i // 5) + 1 for i in range(len(f.trace))])
 
+    def test_traceaccess_from_array(self):
+        a = np.zeros(10, dtype = np.int)
```
pgdr: can you test for `np.int32` and `np.int64`?  Then this looks very good and ready to merge.

jokva:
push'd

PR #115, opened 2017-03-27
Make generator read-ops exception safe

jokva:
Instead of reading directly from file into arrays or buffers that might
already be given to users, read into secondary arrays. This way, when a
read operation fails, user arrays will not be corrupted.


jokva:
Alright, so this should be ready now.

python/segyio/_header.py
```diff
@@ -54,11 +57,14 @@ def __repr__(self):
 
     def readfn(self, t0, length, stride, buf):
```
kjellkongsvik: buf is never used

PR #116, opened 2017-03-27
Implement line_indices with field_forall

jokva:
Turns out the line_indices function was an early version of the
field_forall function, so it might as well be an alias for it. It reads
less from every trace, so it should be faster in most cases.


jokva:
Approve please.

PR #117, opened 2017-03-28
cppcheck support

jokva:
Gonna be some noise here (sorry in advance), so please turn off notifications.


jokva:
This can be reviewed now, but the tests will fail because there's no precise package for cppcheck 1.77. I'll restart the CI once the package is ready.

jokva:
Packages for precise are up, so tests should pass now.

pgdr:
I don't know if we want that for this project, but it _is possible_ to run Travis with Ubuntu 14.04.

jokva:
I suppose 12.-series compatibility is nice, but apparently using trusty sometimes give rise to other problems. A previous iteration of this patch ran on trusty, but now we can run on both trusty and precise.

jokva:
Thanks, much appreciated.

PR #119, opened 2017-03-30
Add examples to README file

mrava87:


pgdr:
Hi, and thanks for the contribution!  If you annotate which language is used for your code snippets in markdown, you get appropriate syntax highlighting.

```
    ```python
    def myfun(no, args):
        do_something();
        assert no is not None
    ```
```


Compare
```
def myfun(no, args):
   do_something();
   assert no is not None
```

with

```python
def myfun(no, args):
   do_something();
   assert no is not None
```

mrava87:
Hi Pål,
that's very useful, I didn't know. I will use it from now on when I add things to documentation 😁

Cheers
iMR

mrava87:
Btw, shall I do it and make new merge request or you adding them for this one?

iMR

pgdr:
You can just add them to this one as a new commit.

mrava87:
Done now, let me know if this is ok. Apparently matlab annotation is not supported by markdown so I only did it for python

PR #145, opened 2017-08-04
WIP: Add test for segyio-cath

haavarop:
Add tests for segyio-cath.

### DIFF
--- a/applications/test/test_segyh.py
+++ b/applications/test/test_segyh.py
@@ -1,0 +1,56 @@
+import unittest
+import subprocess
+
+class TestSegyH(unittest.TestCase):
+
+    work_dir = '../../build'
+    cmd_base = './applications/segyio-cath'
+
+    actual = [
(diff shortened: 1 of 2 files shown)


applications/test/test_segyh.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_help(self):
+        cmd = self.cmd_base + ' --help'
+        result = subprocess.check_output(cmd, shell=True, cwd=self.work_dir)
+        self.assertEqual(result[0:18], b'Usage: segyio-cath')
+
+    def test_version(self):
+        cmd = self.cmd_base + ' --version'
+        result = subprocess.check_output(cmd, shell=True, cwd=self.work_dir)
+        self.assertEqual(result[-5:-2], b'1.2')
```
jokva: Checking a hard-coded segyio-version is a no-no, as this will break on every update.

Besides, the test is rather rigid anyway (and breaks if we should add translation) - we really don't have to check for matching output, just some output really.

jokva: This goes for all the tests, save the actual header output.

PR #164, opened 2017-10-16
Implement segy_ftell with ftello

jokva:
Change the ftell implementation to use the ftello posix function ftello,
enabling 64-bit files on all systems, in partiuclar when long is a
32-bit and regular ftell won't work.

The order of the fstat/ftell macros are changed, preferring
non-underscore prefix'd function when available.


jokva:
I cross-compiled this for armel on debian stretch, and it seems to work nicely.

PR #177, opened 2017-11-09
Tarball dist more robust

jokva:


jokva:
Merged via command line.

PR #189, opened 2017-11-30
Upgrade multibuild to latest devel

jokva:
WIP/test new multibuild


jokva:
Travis issues are resolved and this is no longer relevant.

Issue #201, opened 2018-01-03
error handling using metadata with create


lars-petter-hauge:
Improve error handling/message when combining segyio.tools.metadata with segyio.create

```
spec = segyio.tools.metadata('src_file.segy')
segyio.create('new_file.segy', spec)
new_file = segyio.open('new_file.segy')
```

traceback:
```
     78     try:
---> 79         metrics = segyio._segyio.init_metrics(f.xfd, f.bin.buf)
     80 
     81         f._tr0 = metrics['trace0']

RuntimeError: Input arguments are invalid.
```

jokva:
#206 addresses this by improving the error message. The failure itself is catastrophic and cannot be recovered from.

PR #206, opened 2018-01-04
Improve error message when bin-header is broken

jokva:
The exception raised when the global binary header couldn't be parsed
was less than helpful. The new exception gives a suggestion on what's
wrong.


jokva:
Closes #201

Issue #202, opened 2018-01-03
error message opening incomplete segy file in metadata


lars-petter-hauge:
Given the setup in https://github.com/Statoil/segyio/issues/201
```
spec = segyio.tools.metadata('new_file.segy')
```
Traceback:
```
OSError                                   Traceback (most recent call last)
<ipython-input-2-8128876341bf> in <module>()
----> 1 spec = segyio.tools.metadata('rand_segy.segy')

~/.local/lib/python3.6/site-packages/segyio/tools.py in metadata(f)
    204 
    205     if not isinstance(f, segyio.SegyFile):
--> 206         with segyio.open(f) as fl:
    207             return metadata(fl)
    208 

~/.local/lib/python3.6/site-packages/segyio/open.py in open(filename, mode, iline, xline, strict, ignore_geometry)
     85         f._ext_headers = (f._tr0 - 3600) // 3200  # should probably be from C
     86 
---> 87         dt = segyio.tools.dt(f, fallback_dt = 4000.0) / 1000.0
     88         t0 = f.header[0][segyio.TraceField.DelayRecordingTime]
     89         samples = metrics['sample_count']

~/.local/lib/python3.6/site-packages/segyio/tools.py in dt(segyfile, fallback_dt)
     12     :rtype: float
     13     """
---> 14     return segyio._segyio.get_dt(segyfile.xfd, fallback_dt)
     15 
     16 

OSError: [Errno 0] Error
```

jokva:
Closed by #207

PR #207, opened 2018-01-05
Clarify what fails in segyio.tools.dt

jokva:
The exception messages when segyio.tools.dt fails were less than
helpful, and essentially a direct dump of errno after a failed fread
call, which isn't very helpful, especially when a file is corrupted.

Clarify the exception by making it runtime error (since the file was
already open) and saying whether it was binary- or trace header reading
that failed.

Considering the tiny test program:
```
segyio.tools.metadata(filename)
with segyio.create("foo.sgy", spec) as g, segyio.open(small.sgy) as f:
    zeros = np.zeros([len(f.trace), len(f.trace[0])], dtype = np.single)
    segyio.trace = zeros
segyio.tools.metadata("foo.sgy")
```

Without this patch: IOError: [Errno 0] Error
With this patch:    RuntimeError: Error reading trace header (index 0)

### DIFF
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -621,9 +621,31 @@ static PyObject *py_get_dt(PyObject *self, PyObject *args) {
 
     float dt;
     int error = segy_sample_interval(p_FILE, fallback, &dt);
-    if (error != 0) { return py_handle_segy_error(error, errno); }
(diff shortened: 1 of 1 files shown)


jokva:
Addresses https://github.com/Statoil/segyio/issues/202

PR #236, opened 2018-03-14
Create tracecount arg not in header

jokva:
Fixes https://github.com/Statoil/segyio/issues/235


jokva:
This need a test before it can be merged.

Issue #238, opened 2018-03-15
Support for others data sample format


brunodoamaral:
Hi,

First of all, thanks for sharing this lib. It is a very nice tool for reading seg-y files.

I'd like to help in getting the lib to work with other data sample formats. I was planning to use a public survey to show some seg-y features. I choose to use the [Netherlands Offshore F3 Block - Complete](https://www.opendtect.org/osr/Main/NetherlandsOffshoreF3BlockComplete4GB) data, but once I tried to open its raw seg-y, it gaves me the error `RuntimeError: trace count inconsistent with file size, trace lengths possibly of non-uniform`.

Digging into the issue, I discovered the snippet bellow:

https://github.com/Statoil/segyio/blob/91c46193ffabf586773d1d8125888dbc20c19985/lib/src/segy.c#L712-L716

It is assumed that all trace sample is 4-bytes, but the one I was using is 2-bytes.

I understand that this function should be updated to receive the format, as well as the number of samples, in order to correctly calculate the `trace_bsize`.

I think that this is the 'easy' part, as there aren't a lot of references to this function on the code. However, it seems that treating each format in code should be the hardest part. As this one:

https://github.com/Statoil/segyio/blob/4f1fcfdaaa25d6e608b1d47e2f175f9a81f52cdc/applications/segyinfo.c#L102-L140

Here, one assumes that `trbuf` is a float number and calculate the maximum of trace value. If, instead, the trace is a 2-byte int, than the result will be noise.

I opened this ticket to start a discussion regarding the possible solutions to the issue *and* if the maintainers think that this fix is worth.  

Best regards!
Bruno

jokva:
Hi,

It's certainly possible to support other stuff than 4-byte floats, but I'm not sure if we'll be able to do it without breaking interfaces. Not accepting a variable-width format is one of the bigger regrets from the early design phase, it should've been in the interface from the get-go.

I'll have a quick look over the source code and see if I can figure out exactly what it takes to be able to nicely support non-4-byte inputs.

jokva:
From a *very* quick glance it doesn't seem to be that difficult to support more formats, although it might be slightly inelegant to do it in a way that doesn't break any interfaces. I'll look a bit more into it.

brunodoamaral:
Hi @jokva! I spend some time today in order to make a proof-of-concept implementation of multiple sample format.
[...]

PR #241, opened 2018-03-26
Support for all numeric formats

jokva:
Addresses https://github.com/Statoil/segyio/issues/238


jokva:
For what it's worth, segyio is still broken on systems that use IBM float.

applications/segyio-catr.c
```diff
@@ -295,23 +295,27 @@ static const char* segynames[91] = {
...
+"-s,  --strict                fail on unreadable tracefields\n"
+"-S,  --non-strict            don't fail on unreadable tracefields\n"
+"                             this is the default behaviour\n"
+"-n,  --segyio-names          print with segyio tracefield names\n"
+"-v,  --verbose               increase verbosity\n"
+"     --version               output version information and exit\n"
+"     --help                  display this help and exit\n"
+"\n"
+
+"the -r flag can takes up to three values: start, stop, step\n"
```
lars-petter-hauge: "can take"

lib/include/segyio/segy.h
```diff
@@ -78,9 +94,15 @@ int segy_field_forall( segy_file*,
 
 /*
  * exception: segy_trace_bsize computes the size of the traces in bytes. Cannot
- * fail.
+ * fail. Equivalent to segy_trsize(SEGY_IBM_FLOAT_4_BYTE, samples);
  */
 int segy_trace_bsize( int samples );
+/*
+ * segy_trsize computes the sice of a trace in bytes, determined by the trace
```
lars-petter-hauge: size

lib/src/segy.c
```diff
@@ -711,8 +739,13 @@ int segy_samples( const char* binheader ) {
 
 int segy_trace_bsize( int samples ) {
```
lars-petter-hauge: I don't find it intuitive to have both "segy_trace_bsize" and "segy_trsize", and the difference of the functions by just looking at the names?

lars-petter-hauge: "By default, for backwards compatilibity and convenience, 4-byte samples
are assumed." <-- backwards compatibility is the reason?

jokva: Yes, backwards compatibility is the sole reason, otherwise I'd reuse the trace_bsize name. This way, old code keeps working as it has, and only new code targeting non-IBM float formats needs to change.

jokva:
@brunodoamaral This is shaping up now. It'd be great if you could test out this patch series and see if it works as expected for you.

jokva:
I'm merging this now. Tests pass, and the best way to get some real-world testing on it is to integrate it in master and maybe even do a binary alpha release.

There's a few more features planned for 1.6, so the actually release of 1.6 is probably a month or two away.

Closes #238

bluetyson:
I got this : RuntimeError: trace count inconsistent with file size, trace lengths possibly of non-uniform
today too with segyio in python, so I appreciate the work on this.

jokva:
What kind of file gave you this error?

Khalilsqu:
> 
> 
> I got this : RuntimeError: trace count inconsistent with file size, trace lengths possibly of non-uniform
> today too with segyio in python, so I appreciate the work on this.

I am also  getting the same error?? file type is `sgy`

I can read it with 

[SeiSee](http://www.dmng.ru/en/freeware.html)

image

jokva:
Hi,

The 8-byte floating-point format was new with revision 2. Supporting it shouldn't be too difficult though, and is a nice feature for 1.9

Just to verify, what's the value in bytes 3225-3226 in the binary header?

jokva:
Please continue the discussion in issue #353

PR #244, opened 2018-04-02
Sphinx config + readthedocs

jokva:


jokva:
This is shaping up. What's currently lacking is properly configured readthedocs that freezes on tags and builds latest.

PR #248, opened 2018-04-03
Write sample-count per trace

jokva:
Update the sample count per trace whenever cropping a file.


applications/segyio-catr.c
```diff
@@ -474,6 +474,7 @@ int main( int argc, char** argv ) {
         default: labels = su; break;
     }
 
+    /* verify all ranges are proper */
```
lars-petter-hauge: Should this end in something like "..rly defined"..?

jokva: I suppose that would be better English, sure.

jokva: The ranges would always be defined, but they might be horribly broken.

PR #259, opened 2018-05-18
Refactor python protocol support, add better docstrings

jokva:
PR to make travis do its job.


lib/src/segy.c
```diff
@@ -896,8 +896,8 @@ int segy_sample_interval( segy_file* fp, float fallback, float* dt ) {
         return err;
     }
 
-    int bindt = 0;
-    int trdt = 0;
+    int32_t bindt = 0;
+    int32_t trdt = 0;
```
JensGM: Is this because the format is guaranteed to use 32 bit ints?

jokva: Yes, it's specified as up-to 4-byte two's complement integers. Most are 2-byte ints, but some are 4.

jokva:
This is finally shaping up. The change set is huge, but feature-wise not too much happens.

# Synopsis
- Files have been moved a bit around, but in total there are fewer. Some classes are grouped in files based on their behaviour (trace and header, line and header-line)
- Files prefixed with `_` are largely gone. It's still not a public feature to construct the individual classes, but hiding their names won't do much.
- Docstrings are written and concentrated on `__getitem__`, `__setitem__` etc., instead of all being in segy.py. This has been accounted for in the sphinx documents too.
- Support for protocols has been formalised through the `collections` module, which cleans up some interfaces and adds support for a few new features such as `index()` and `x in c`.
- Most `__getitem__` etc are rewritten to a much simpler style, to make maintenance easier. This is the largest change of this patch series, but in effect, operations behave better and more predictably.

PR #271, opened 2018-08-28
Use fixture + test-case for set-format test

jokva:
Move from the BDD-section to TEST_CASE_METHOD. The goal is to make the
actual test shorter and more to the point, at the cost of slightly more
convoluted dependency chain.

### DIFF
--- a/lib/test/segy.cpp
+++ b/lib/test/segy.cpp
@@ -219,6 +219,25 @@ struct smallbin : smallfix {
     }
 };
 
+struct smallbasic : smallfix {
+    static const long trace0 = 3600;
+    static const int trace_bsize = 50 * 4;
+};
+
+bool success( Err err ) {
+    return err == Err::ok();
+}
+
+int arbitrary_int() {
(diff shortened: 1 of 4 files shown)


jokva:
Opinions please.

lib/test/segy.cpp
```diff
@@ -192,6 +168,151 @@ void regular_geometry( segy_file* fp,
...
+                  MMAP_TAG "crossline sorting is detected with swapped il/xl",
+                  MMAP_TAG "[c.segy]" ) {
+    int sorting;
+    Err err = segy_sorting( fp, xl, il, of, &sorting, trace0, trace_bsize );
+    CHECK( success( err ) );
+    CHECK( sorting == SEGY_CROSSLINE_SORTING );
+}
+
+TEST_CASE_METHOD( smallfields,
+                  MMAP_TAG "invalid byte-offsets are detected",
```
JensGM: This is not what is tested

PR #275, opened 2018-09-03
 Fix an issue where segyio-catr printed the same trace twice

ErlendHaa:
While segycatr takes range arguments as 1-indexed, it would still
print the first trace if 0 was given as --range or -trace. Hence, situasions
where the first trace was printed twice would occur. Now, zero is no longer
accepted as a vaule for --range or --trace. This behavior is added to man-page

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -997,15 +997,18 @@ int segy_sorting( segy_file* fp,
     err = segy_traceheader( fp, 0, traceheader, trace0, trace_bsize );
     if( err != SEGY_OK ) return err;
 
-    if( il < 0 || il >= SEGY_TRACE_HEADER_SIZE )
-        return SEGY_INVALID_FIELD;
-
-    if( xl < 0 || xl >= SEGY_TRACE_HEADER_SIZE )
-        return SEGY_INVALID_FIELD;
-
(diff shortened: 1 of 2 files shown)


jokva:
```$ git log -1 35078e0d0c694507f8573a9882ccbba15cc0a1e9
commit 35078e0d0c694507f8573a9882ccbba15cc0a1e9
Author: Jørgen Kvalsvik <jokva@statoil.com>
Date:   Tue Apr 3 10:10:16 2018 +0200

    Handle default properly in segyio-catr
    
    By default, catr prints the first trace header, relying on calloc
    setting the range-field to zero. This wasn't checked for later though,
    and the main program loop expectets 1-index'd ranges.
    
    In the case of defaults, set the start field to 1
```

So the intention is 1-indexing, which at least makes the man-page correct.

jokva:
I think actually just the `if 0: continue` check in the range-sanitiser should be fine, unless you know of a case that would fall through.

That being said, `segyio-catr -t 1 -t 1 -t 1` would print the first trace 3 times, and in general multiple partially overlapping ranges would lead to the same trace being printed multiple times. A better `segyio-catr` would normalise the input ranges and be unique in output. I'll ake

Issue #217, opened 2018-01-18
f.text does not support indexing or slicing


jokva:
Although not that commonly used, the text headers can be considered an list-of-headers and, for consistency, should probably support these operations.

jokva:
Closed by #283

PR #283, opened 2018-09-10
Make attributes into Sequence

jokva:
Move the implementation of the f.attributes implementations into
trace.py, and make it a proper Sequence.

Attributes is probably the most obvious array_like of all the segyio
addressing modes, so it not being a sequence was an oversight at best.


jokva:
Closes #217

PR #287, opened 2018-09-18
Two phase init python fd

jokva:
Use two-phase init for python file handle

Split the python file handle initialisation routine into two phases -
one for the physical, on-disk open, and one for filling out meta-data.
In short, instead of the previous interface of:
```
    fd = segyio.SegyFile(...)
```
it's now necessary to do:
```
    fd = _segyio.segyiofd(...)
    fd.segyopen()

    fd = _segyio.segyiofd(...)
    fd.segymake(...)
```
for opening and making files, respectively. This does not leak into user
code, and is taken care of inside open and create.

This properly splits the open/create routines into separate functions,
instead of interleaving them with conditionals. This is a readability
win, at the cost of some minor duplication.

### DIFF
--- a/python/segyio/create.py
+++ b/python/segyio/create.py
@@ -168,17 +168,21 @@ def create(filename, spec):
     ext_headers = spec.ext_headers if hasattr(spec, 'ext_headers') else 0
     samples = numpy.asarray(spec.samples)
 
-    binary = bytearray(_segyio.binsize())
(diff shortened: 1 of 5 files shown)


JensGM:
Would those details leak out to user code originally though?

jokva:
No, this is all contained inside the library. No user code is affected.

PR #289, opened 2018-09-24
Call write, not read, in mmap write-trace

jokva:
The wrong function (!) was called here, meaning memeory mapped writes
never really worked as expected. This was not caught by a test, because
the tests for this function use new files, and memory freshly-created
files has never been supported.

The write-sub-trace tests have been amended to also work with
memory-mapping. The new strategy for these tests to make a test-specific
copy of the small.sgy file, and open it in read-write mode, instead of
creating a new truncated test-specific file as before.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -1473,8 +1473,8 @@ int segy_writesubtr( segy_file* fp,
          * be handled by the stride-aware code path
          */
         if( fp->addr ) {
-            err = memread( fp, fp->cur, buf, elemsize * elems );
-            if( err != SEGY_OK ) return err;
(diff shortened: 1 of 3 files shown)


lib/src/segy.c
```diff
@@ -1473,8 +1473,8 @@ int segy_writesubtr( segy_file* fp,
          * be handled by the stride-aware code path
          */
         if( fp->addr ) {
-            err = memread( fp, fp->cur, buf, elemsize * elems );
-            if( err != SEGY_OK ) return err;
+            err = memwrite( fp, fp->cur, buf, elemsize * elems );
```
JensGM: read replaced with write?

jokva: Yep, pretty embarassing.

PR #299, opened 2018-09-28
cleanup unused variable

ErlendHaa:


jokva:
auditwheel just broke because of some changes to wheel, so this build will remain broken until upstream is fixed, and we pull a new multibuild. This will remain red for a little bit.

https://github.com/matthew-brett/multibuild/issues/202

Issue #293, opened 2018-09-26
Helper function to create complete SEG-Y file from numpy array


jokva:
A  `from_array` or similar that function that takes a 2/3D array and create a SEG-Y with defaulted headers.

For a lot of use cases, this hugely simplifies operations where you don't care too much about correctness, but want your slice or volume to work somewhat nicely with other tools.

I think this will also be a useful development tool, being able to almost declaratively create test files with proper formatting, with little effort.

jokva:
Closed with #304

PR #304, opened 2018-10-11
Helper function to create a SEGY file from a 2D/3D numpy array

ErlendHaa:
Connects to #293

### DIFF
--- a/python/segyio/tools.py
+++ b/python/segyio/tools.py
@@ -1,4 +1,8 @@
 import segyio
+from segyio import TraceField
+from segyio import TraceSortingFormat
+from segyio import SegySampleFormat
+
 import numpy as np
 import textwrap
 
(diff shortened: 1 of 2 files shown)


python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
 
     return f
+
+
+def create_from_array2D(filename, data):
+    """ Create a new SEGY file from a 2D array
+
+    Create an unstructured SEGY file from a 2-dimentional array with defaulted
+    headers. The tracecount and samplecount are inferred from the size of the
+    array. All traceheader fields are defaulted to zero. The file have a
```
jokva: I guess "all" fields aren't zero'd, which should be clear from documentation.

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+        trace n | s0 | s1 | ... | sn |
+                 --------------------
+
+    traces = [0, len(axis(0)]
+    samples = [0, len(axis(1)]
+
+    Parameters
+    ----------
+
+    filename : str
```
jokva: This actually only has to be string-like, i.e. we accept py.path, localpath and friends.

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+        Path to file to create
+    data : 2-dimensional array_like
+
+    Notes
+    -----
+
+    .. version added 1.8
```
jokva: `.. versionadded::` without space.

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+
+    Examples
+    --------
+
+    Create a file from a 2D array, open it and read a trace:
+
+    >>> segyio.tools.creat_from_array(path, array2d)
```
jokva: Typo.

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+    >>> segyio.open(path, mode, strict=False) as f:
+    ...     tr = f.trace[0]
+    """
+    data = np.asarray(data)
+    dimentions = len(data.shape)
+
+    if dimentions != 2:
+        problem = "Invalid array dimentions"
```
jokva: Reads a bit off, maybe `Expected 2 dimensions, was {}` or similar?

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+                             TraceField.CDP_TRACE             : tr,
+                             TraceField.offset                : 1,
+                             TraceField.INLINE_3D             : 1,
+                             TraceField.CROSSLINE_3D          : tr,
+                             TraceField.TRACE_SAMPLE_COUNT    : samplecount,
+                             TraceField.TRACE_SAMPLE_INTERVAL : 4000
+                           }
+        f.bin.update(
+            tsort=2,
+            rev=1
```
jokva: I'm a bit uncomfortable setting `rev=1`, because a compliant revision 1 file should have some more data set that we can't without help.

python/segyio/tools.py
```diff
@@ -402,3 +406,178 @@ def resample(f, rate = None, delay = None, micro = False,
...
+    spec = segyio.spec()
+    spec.format = SegySampleFormat.IBM_FLOAT_4_BYTE
+    spec.ilines = list(range(ilinecount))
+    spec.xlines = list(range(xlinecount))
+    spec.samples = list(range(samplecount))
+    spec.sorting = TraceSortingFormat.INLINE_SORTING
+
+    with segyio.create(filename, spec) as f:
+        for ilno in spec.ilines:
+            iline = np.concatenate([data[ilno,xl,:] for xl in spec.xlines])
```
jokva: *Nice*

jokva:
There's a few typos still (`dimentions`), and I'd appreciate if you'd compact the docstrings a bit.

Does sphinx run cleanly? `cmake -DBUILD_DOC=ON && make doc`

python/segyio/tools.py
```diff
@@ -403,3 +407,429 @@ def resample(f, rate = None, delay = None, micro = False,
     f._samples = (np.arange(len(f.samples)) * rate) + t0
 
     return f
+
+
+def from_array(filename, data, iline=189,
+                               xline=193,
+                               sample_format=SegySampleFormat.IBM_FLOAT_4_BYTE,
+                               dt=4000,
+                               delay_recording_time=0):
```
jokva: Personally, I think `delay_recording_time` and also to some extent `samle_format` is quite the mouthful. I guess `sample_format` is ok, but what do you think of the (SU-derived) `delrt` for delay-recording-time?

python/segyio/tools.py
```diff
@@ -403,3 +407,429 @@ def resample(f, rate = None, delay = None, micro = False,
...
+    """
+
+    data = np.asarray(data)
+    dimentions = len(data.shape)
+
+    if dimentions not in range(2,5):
+        problem = "Expected 2, 3 or 4 dimensions, {} was given".format(dimentions)
+        raise ValueError(problem)
+
+    if dimentions is 2:
```
jokva: While cute and probably ok, you actually want `==` here, since integer-constants probably won't be the same object as an int-var

python/segyio/tools.py
ErlendHaa: I agree on using the `deltr`. What do you think of using just `format`. Although it redefines the built-in `format`, it should not cause any confusion due to the different nature of the two!?

Issue #294, opened 2018-09-26
Method to (re)interpret structure on top of an open file


jokva:
`f.interpret` should take a set of structure-defining parameters, and populate internal structures such as inlines, crosslines, sorting etc. to enforce a known structure that isn't properly backed by headers, or add structure where there is none.

This can be both an escape hatch and a useful tool for creative interpretation of volumes.

jokva:
#311 does most of this, but lacks for the new `su.open` and the `create` function.

ErlendHaa:
As `interpret` is called from `infer_geometry`, #311 does in fact support `su.open` as well. I will make a new PR where `create` uses `interpret`

jokva:
Mea culpa. Just create then.

PR #311, opened 2018-10-19
Method for (re)interpret structure on an open file

ErlendHaa:
Given a set off structure-defining parameters interpret() will
(re)interpret the structure of the file. This does not change the
file itself, it only changes segyio's interpretation of it.
Connects to #294

### DIFF
--- a/python/segyio/segy.py
+++ b/python/segyio/segy.py
@@ -838,6 +838,114 @@ class SegyFile(object):
 
         return '+' not in self._mode
 
+
+    def interpret(self, ilines, xlines, offsets=None, sorting=TraceSortingFormat.INLINE_SORTING):
+
+        """ (Re)interpret structure on top of a file
+
+        (Re)interpret the structure of the file given the new sorting, ilines,
+        xlines and offset indices. Note that file itself is not changed in any
+        way, it is only segyio's interpretation of the file that changes.
+
+        interpret() expect that the ilines-, xlines- and offsets-indices are
+        unique. It also expect the dimensions of ilines, xlines and offset to
+        match the tracecount.
+
+        Parameters
+        ----------
+
+        f : SegyFile
+        ilines : array_like
+            ilines indices in new structure
+        xlines : array_like
+            xlines indices in new structure
+        offsets : array_like
+            offset indices in new structure
+        sorting : int, string or TraceSortingFormat
+            Sorting in new structure
+
+        Notes
+        -----
+
+        .. versionadded:: 1.8
+
+        Examples
+        --------
+        (Re)interpret the structure of the file:
+        >>> ilines = [10, 11, 12, 13]
+        >>> xlines = [20, 21, 22, 23, 24]
+        >>> with segyio.open(file, ignore_geometry=True) as f:
+        ... f.interpret(ilines, xlines)
+        """
+
+        valid_sortings = {
+            1           : TraceSortingFormat.CROSSLINE_SORTING,
+            2           : TraceSortingFormat.INLINE_SORTING,
+            'iline'     : TraceSortingFormat.INLINE_SORTING,
+            'inline'    : TraceSortingFormat.INLINE_SORTING,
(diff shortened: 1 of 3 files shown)


ErlendHaa:
Codacy complains about the danger of using array as default value:

> Dangerous default value [] as argument
To a Python novice, this may seem like a reasonable way to default to an empty array for the extras param. What really happens is that this "default" array gets created as a persistent object, and every invocation of my_method that doesn't specify an extras param will be using that same list object—any changes to it will persist and be carried to every other invocation!

However, in this case the variable in question is never changed within the function so do we really care?

jokva:
> However, in this case the variable in question is never changed within the function so do we really care?

Yes. Not because there's a problem now, but because it's *very* easy to make a mistake in the future. The `offsets = offsets or [0]` or `if offsets is None: offsets = [0]` idiom is common enough (and cheap enough) to warrant using it over easy-to-break code.

python/segyio/segy.py
```diff
@@ -833,6 +833,83 @@ def readonly(self):
 
         return '+' not in self._mode
 
+
+    def interpret(self, ilines, xlines, offsets=None, sorting=2):
```
jokva: `sorting` should either be None or the enum, I gues, for free documentation.

python/segyio/segy.py
```diff
@@ -833,6 +833,83 @@ def readonly(self):
...
+        f : SegyFile
+        ilines : array_like
+            ilines indecies in new structure
+        xlines : array_like
+            xlines indecies in new structure
+        offset : array_like
+            offset indecies in new structure
+        sorting : int
+            Sorting in new structure
+
```
jokva: Some spelling issues in this entire section. indecies -> indices, and I guess the description could use some reflow. 

`sorting` should be clear on accepting either an enum or maybe even a string: `f.interpret( ..., sorting = 'inline')`

jokva: Could also clarify that `segyio.open(file) == segyio.open(ignore_geometry = True).interpret(...)`

python/segyio/segy.py
```diff
@@ -833,6 +833,83 @@ def readonly(self):
...
+        >>> ilines = [10, 11, 12, 13]
+        >>> xlines = [20, 21, 22, 23, 24]
+        >>> with segyio.open(file, ignore_geometry=True) as f:
+        ... f.interpret(ilines, xlines)
+        """
+
+        if offsets is None:
+            offsets = np.arange(1)
+
+        ilines = np.asarray(ilines)
```
jokva: I *think* this also needs a `dtype = int` hint.

python/segyio/segy.py
```diff
@@ -833,6 +833,83 @@ def readonly(self):
...
+        xlines = np.asarray(xlines)
+        offsets = np.asarray(offsets)
+
+        if np.size(ilines) * np.size(xlines) * np.size(offsets) != self.tracecount:
+            error = "The length of ilines * xlines * offsets does not match the number of traces"
+            raise ValueError(error)
```
jokva: Maybe add some info on dimensions here? msg + solution style, since this is fairly user facing

python/segyio/segy.py
ErlendHaa: Also check that int coercion does not introduce duplicates

python/segyio/segy.py
```diff
@@ -833,6 +833,83 @@ def readonly(self):
...
+            raise ValueError('{}, {}'.format(error, solution))
+
+        from . import _segyio
+
+        self._sorting = sorting
+        self._offsets = np.copy(np.asarray(offsets, dtype=np.intc))
+        self._ilines = np.copy(np.asarray(ilines, dtype=np.intc))
+        self._xlines = np.copy(np.asarray(xlines, dtype=np.intc))
+
+        line_metrics = _segyio.line_metrics(self.sorting,
```
ErlendHaa: Do this first to provide more exception safety

PR #328, opened 2018-11-22
Tidying up imports

achaikou:
Pycharm doesn't complain about the code, with some additional setup able to run tests from IDE.
Worked on 2.7, IDE and ctest, waiting for other versions confirmation

### DIFF
--- a/python/segyio/binfield.py
+++ b/python/segyio/binfield.py
@@ -1,5 +1,4 @@
-from segyio import Enum
-
+from . import Enum
 
 class BinField(Enum):
     """Trace header field enumerator
--- a/python/segyio/segy.py
+++ b/python/segyio/segy.py
@@ -11,7 +11,7 @@ from .line import Line
 from .trace import Trace, Header, Attributes, Text
 from .field import Field
 
-from segyio.tracesortingformat import TraceSortingFormat
+from .tracesortingformat import TraceSortingFormat
 
 
 
--- a/python/segyio/segysampleformat.py
+++ b/python/segyio/segysampleformat.py
@@ -1,5 +1,4 @@
-from segyio import Enum
-
+from . import Enum
 
 class SegySampleFormat(Enum):
     IBM_FLOAT_4_BYTE = 1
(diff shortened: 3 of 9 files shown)


python/segyio/line.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, filehandle, labels, length, stride, offsets, name):
         self.dtype = filehandle.dtype
 
         # pre-compute all line beginnings
-        from ._segyio import fread_trace0
+        from segyio._segyio import fread_trace0
```
jokva: Isn't this the other way around?

achaikou: Code was changed the way to make Pycharm happy
Pycharm complains about using ._segyio here as I think it cannot see .so file in '.' as it's not there.
When doing segyio._segyio and having build/python folder on interpreter path it doesn't complain as per my understanding it searches for .so file in segyio package and finds it under build folder. 
My intent was to use . where possible to use local files, and leave segyio. only where we should use files from package.

Please let me know if I misunderstood something and using .segyio here would make an actual difference

JensGM: This might also be fixed by importing absolute import from `__future__`?

achaikou: Not sure why this matters? Per my understanding future import decides that it will start looking for packages from top ones like site_packages (and similar) ones, rather than from local .py files.
For source/python/segyio/line.py location (IDE "uresolved reference" point of view) the only '_segyio' it knows about is the build segyio (if on the path), hence able to refer through package only.
From build/python/segyio/line.py (execution point of view) it already has _segyio.so right there in '.', so execution goes fine even though IDE complains.
Per my understanding we do not overwrite any package/module in either case as we only have one - build/python/segyio (by default source/python/segyio isn't on the path)

(+Tried to add absolte future import to the code, see no difference in behaviour with 'segyio.' and '.', running from IDE and ctest, but can't guarantee I don't miss something obvious)

jokva: >  When doing segyio._segyio and having build/python folder on interpreter path it doesn't complain as per my understanding it searches for .so file in segyio package and finds it under build folder.

_segyio will still always be a local file to the module itself though, so I'll accept some complaining from pycharm to preserve the intent we get from `from .`

achaikou: Done. Changed back few other places where used segyio._segyio instead of local '.'

PR #330, opened 2018-11-27
Fix Text.__setitem__ for Text parameter

achaikou:
Update from discussion: my bad, didn't notice that if slice is provided, list (not one string) is expected as val . Updated the code to take first element from each provided Text.

### DIFF
--- a/python/segyio/trace.py
+++ b/python/segyio/trace.py
@@ -1,6 +1,7 @@
 import collections
 import contextlib
 import itertools
+import warnings
 import sys
 try: from future_builtins import zip
 except ImportError: pass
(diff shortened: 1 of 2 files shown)


python/test/segy.py
```diff
@@ -558,6 +559,10 @@ def test_read_text_sequence():
             assert text
 
         assert iter(f.text)
+        with warnings.catch_warnings(record=True) as w:
```
jokva: This is much simple with pytest (I think):

https://docs.pytest.org/en/latest/warnings.html#ensuring-code-triggers-a-deprecation-warning

python/segyio/trace.py
```diff
@@ -842,9 +843,26 @@ def __setitem__(self, i, val):
 
         >>> f.text[1] = g.text[0]
 
+        If i is slice [i_1:i_n] and val=[val_1, .. val_n],
```
jokva: This documentation is very confusing to me.

In `__setitem__` for trace headers, the following is written:

> If i or o is a slice, val must be an iterable.

This _very_ briefly describes the behaviour. This is then made obvious by the actual examples.

python/segyio/trace.py
```diff
@@ -842,9 +843,26 @@ def __setitem__(self, i, val):
...
+
+        >>> f.text[1] = g.text
+        >>> assert f.text[1] == g.text[0]
+
+        If i is slice [i_1:i_n] and val is [Text_1, ... Text_n],
+        text[k] = val[k][0] for k in 1..n
+
+        >>> f.text[1:3] = [g1.text, g2.text]
+        >>> assert f.text[1] == g1.text[0]
+        >>> assert f.text[2] == g2.text[0]
```
jokva: Please review the numpy docstring format - this entire section will probably not render well.

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

You can build the documentation locally by building with `cmake -DBUILD_DOC=ON` and running `make doc`

PR #332, opened 2018-11-28
[WIP] Segyio sort

### LINKED ISSUE #326: Detect descending line numbers as sorted
See #322 for a case leading up to this.

Currently, segyio only considers a file sorted if the line numbers are _ascending_. Ideally, as long as the line numbers move regularly, ascending and descending shouldn't matter, and the file from #322 should be detected as sorted.

ShikhaMishra9:
Connects to #326 
Closes #326


lib/test/segy.cpp
```diff
@@ -1172,6 +1172,63 @@ SCENARIO( "modifying trace header", "[c.segy]" ) {
...
+                            SEGY_TR_OFFSET,
+                            &sorting,
+                            trace0,
+                            trace_bsize );
+    CHECK( success( err ) );
+    segy_close( fp );
+    return sorting;
+}
+
+TEST_CASE("", "[c.segy]") {
```
jokva: These all need names.

lib/test/segy.cpp
```diff
@@ -1172,6 +1172,63 @@ SCENARIO( "modifying trace header", "[c.segy]" ) {
...
+
+    Err err = segy_sorting( fp,
+                            SEGY_TR_INLINE,
+                            SEGY_TR_CROSSLINE,
+                            SEGY_TR_OFFSET,
+                            &sorting,
+                            trace0,
+                            trace_bsize );
+    CHECK( success( err ) );
+    segy_close( fp );
```
jokva: If you use the unique_ptr thing around openfile, you don't need an explicit close, and you remove the possibility for a leak here,

python/examples/sorting-permutation.py
```diff
@@ -0,0 +1,73 @@
...
+
+        with segyio.open(sourcefile) as src, segyio.open(fname, 'r+') as dst:
+            for i in range(1 + src.ext_headers):
+                dst.text[i] = src.text[i]
+
+            dst.bin = src.bin
+            dst.trace = src.trace
+            dst.header = src.header
+
+    with segyio.open(sourcefile) as src, segyio.open(pathjoin("dec-il-xl-off", destfile), 'r+') as dst:
```
jokva: This is too long and unwieldy for my taste, and I think you even benefit readability-wise from doing it in two blocks.

python/examples/sorting-permutation.py
```diff
@@ -0,0 +1,73 @@
+import sys
+import os
+import shutil
+import segyio
+
+
+def pathjoin( prefix, path):
```
jokva: Consistent spacing please.

jokva:
Merged in https://github.com/Statoil/segyio/commit/29574259a48589c31181fc6d3f0467d65c907188

PR #344, opened 2018-12-19
Add tests for field_forall and field_foreach

### LINKED ISSUE #224: f.attributes could use some tests
`f.attributes`, the `field_foreach` and `field_forall` internal functions could use some tests.

achaikou:
Closes #224 

Upper level: no additional tests for f.attributes. There already is a test method with good coverage (test_attributes(openfn, kwargs)), don't think anything else is useful 
Lower level: actual C method (segy_field_forall) is tested as well.
Hence adding only 'middleware' tests similar to existing ones.
P.S. Wanted to add some bad-index test for foreach, but apparently it's not that easy to catch SIGSEGV. Not sure if we shouldn't have some additional checks in C code to provide users with better error message. Leaving as is for now.

### DIFF
--- a/python/test/segyio_c.py
+++ b/python/test/segyio_c.py
@@ -427,6 +427,49 @@ def read_and_write_traceheader(f, mmap):
     f.close()
 
 
+def test_read_traceheaders():
+    f = _segyio.segyiofd("test-data/small.sgy", "r", 0).segyopen()
+    read_traceheaders_forall(f, False)
+    read_traceheaders_forall(f, True)
+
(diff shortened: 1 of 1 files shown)


jokva:
> P.S. Wanted to add some bad-index test for foreach, but apparently it's not that easy to catch SIGSEGV. Not sure if we shouldn't have some additional checks in C code to provide users with better error message. Leaving as is for now.

That's ok, that's _by design_. It's assumed that kind of checking is done in higher levels, and that once you reach the C core every index is fine.

PR #354, opened 2019-02-25
Fix reading of dt when either binary or trace header value is negative

aimanuslim:
Hi, this may be an issue with not adhering to SEGY files convention, but I am opening this just in case. When I try to read the dt value using `header[segyio.su.dt]` , I am getting an incorrect value for the case when either one of the dt values in binary and trace header is negative.

For example,

Binary dt value = -25565
Trace dt value = 40000

The correct dt value should be 40000 however, due to this part of the code in `segy.c`

```
    /*
     * 3 cases:
     * * When the trace header and binary header disagree on a (non-zero)
     *   sample interval; choose neither and opt for the fallback.
     * * When both sample intervals are zero: opt for the fallback.
     * * Otherwise, choose the interval from the non-zero header.
     */

    *dt = fallback;
    if( `binary_header_dt == 0 && trace_header_dt != 0` )
        *dt = trace_header_dt;

    if( trace_header_dt == 0 && binary_header_dt != 0 )
        *dt = binary_header_dt;

    if( trace_header_dt == binary_header_dt && trace_header_dt != 0 )
        *dt = trace_header_dt;

    return SEGY_OK;
}
```

the dt value that will be chosen is the fallback value, which somehow is the binary header value of my segy file. In order to prevent this, I am providing the fix here. I havent tried compiling it again yet due to constraint on resources. You can edit/reject the fix if you think the fix is unnecessary.


jokva:
I think the patch looks fine, and indeed negative values in the bin-fields really don't make sense.

The only issue I have with it is the same reason the original code is quietly problematic - SEG-Y defines no "null" or "missing" value, probably because in a lot of fields 0 is a valid value. Our _guess_ was that for most files (which conformed with what we'd seen at the time) 0 was used as a null value, and we wanted the fallback to happen fairly fast.

Rambling aside, I don't see too much harm happening here - it's simple enough to inject your own sample interval anyway.

jokva:
I took the liberty of pushing directly to your branch, I hope that's ok with you. I expanded the commit message a little bit, and added a test for this new behaviour..

Once the test suite runs travis and appveyor I'll merge this. Thanks!

PR #359, opened 2019-03-05
More numerical formats

### LINKED ISSUE #353: Support for more numerical formats
Segyio could quite easily support several of the numerical formats introduced with SEG-Y rev2. Among them are 8-byte integer and 8-byte IEEE float.

jokva:
Add support for more numeric formats, most of them introduced with SEG-Y
rev2:

- 8-byte IEEE float (double)
- 1-byte signed and unsigned integer (char)
- 2-byte signed and unsigned integer (short)
- 4-byte signed and unsigned integer (int)

Some prep work for 3-byte integers is done, but since neither C nor
numpy has builtin support for that, it's not available in segyio still.

The code is based on the work of Hassan Sabirin who both proposed the
change, and did an implementation [1].

[1] #353

Closes #353


hassan-sabirin:
Everything looks sane to me ;)

PR #371, opened 2019-05-05
Add quick start example with plot

hallgrim:
The quick start section should show a few common usecases, and be clear in terms of usage patterns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,12 +61,31 @@ start your favourite Python interpreter and type `help(segyio)`, which should
 integrate well with IDLE, pycharm and other Python tools.
 
 ### Quick start ###
+
+Two examples on how to use segyio:
+
+Integrate with numpy to filter for a given treshold value
 ```python
 import segyio
 import numpy as np
+threshold = 1e-2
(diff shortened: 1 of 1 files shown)


mrava87:
Hi, 
nice idea :)

Since you also add plots then I would say that this is more appropriate for the companion repo https://github.com/equinor/segyio-notebooks (we should have a link to it  in the README?). In that repo you can make a notebook running those steps and provide the data or a set of commands for downloading the data so that everyone can reproduce your example...

mrava87:
PS: actually realized we already have a link to that repo in README, ``example notebooks`` in this sentence is linked to that repo ``When segyio is built and installed, you're ready to start programming! Check out the tutorial, examples, example programs, and example notebooks``

hallgrim:
The idea of the simple plot was to quickly show off the capabilities of the library, and create a visual version of what it can achieve. I do agree that examples belongs in the notebook companion, in particular examples that pulls in other dependencies. 

We were trying to make the once exception that emphasis what kind of things you can see in the notebooks and that would make sure that it was easy to realise what this library can do.

PR #372, opened 2019-05-08
Added example for a trace resample in doc string

harrymchugh:
Added a one liner to perform a resample of the trace (e.g from 4ms sample rate to 2ms sample rate)

### DIFF
--- a/python/segyio/trace.py
+++ b/python/segyio/trace.py
@@ -84,6 +84,10 @@ class Trace(Sequence):
     >>> tr = tr - 100
     >>> avg = numpy.average(tr)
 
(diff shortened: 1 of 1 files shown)


python/segyio/trace.py
```diff
@@ -84,6 +84,10 @@ class Trace(Sequence):
     >>> tr = tr - 100
     >>> avg = numpy.average(tr)
 
+    Perform some seismic processing on a trace. E.g resample from 2ms spacing to 4ms spacing (note there is no anti-alias filtering in this example):
```
jokva: This line is a bit long - please reflow it to be 79 characters inclusive.

harrymchugh: this has been done with the next commit

jokva: Thank you, I appreciate it.

Issue #340, opened 2018-12-17
Reading subset of a trace from t1 to t2 (or z1 to z2)


mrava87:
Hello,
we would like to read a big (eg prestack) data in memory using the segyio.tools.collect(segyfile.trace[:]) routine. However we are only interested in reading a subset of samples (ie time or depth axis) which fits into memory while reading the entire data for all samples does not. One way could be to make a for loop/list comprehension ourselves but this should perform poorer segyio.tools.collect. We wonder if there is a case here to build something to read a subset of a trace like we do for trace, iline, xline, gather?

jokva:
Addressed in #383, at least for reading.

PR #383, opened 2019-06-02
Feature/trace slicing

pederka:
Cater to the resolving of issue #340 by adding the capability for trace sub-slicing without reading the whole trace into memory. While previously one could do

f.trace[0][10:20]

you can now use

f.trace[0, 10:20]


jokva:
Very nice, good work!

I'm going to merge this now, so that we can eventually release a new alpha of 1.9 so people can try it out and give some feedback. I'll squash the history for you.

jokva:
The control flow is a little bit convoluted, and will probably be revised before the release. Also, maybe it's an idea to support subtrace _writes_ in the future. That's for another day.

pederka:
Thanks! The convoluted control flow favors speed in default cases over structure - but I think it definitely can be improved upon. Thanks for squashing the rather messy branch history.

PR #408, opened 2019-09-25
Allow to have trace containing more than 32767 samples

### LINKED ISSUE #407: Can not open SEGY with long traces
I tried to open a file with traces of 60000 samples each and I got the following error: `RuntimeError: trace count inconsistent with file size, trace lengths possibly of non-uniform`

After some debug I found out that it's because the "Number of expected samples per trace" is read as a signed 16 bit value. Is there any reason for that ?

I understand that the SEGY standard does not explicitly mention that this field is unsigned but a negative number of sample per trace does not really make sense

kurddt:
Read the number of samples as an unsigned 16bits instead of signed 16bits

Resolve #407


kurddt:
I made an other PR #409 which also resolves #407 but also allow to display correctly the unsigned header values

PR #459, opened 2020-04-17
Raplima docupdate 1

raplima:
Hi, I would like to suggest two small changes, both related to the "unable to find sorting" message. One in the README.md, one in the message itself. I know I spent a little longer I would like to admit trying to understand why that was happening a while ago and I know a colleague bumped in the same issue recently. I hope this modification is helpful!

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,6 +604,7 @@ just a collection of traces in arbitrary order, this would fail.
 
 #### Possible solutions
 
+Check if segyio.open `iline` and `xline` input parameters are correct for current file.
 Segyio supports files that are just a collection of traces too, but has to be
 told that it's ok to do so. Pass `strict = False` or `ignore_geometry = True`
 to `segyio.open` to allow or force unstructured mode respectively. Please note
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -841,7 +841,10 @@ struct metrics_errmsg {
                                    "or offset (%i) field", il, xl, of );
 
             case SEGY_INVALID_SORTING:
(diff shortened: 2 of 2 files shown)


jokva:
I'll merge this once the compile error is fixed - you need to add an overload for runtime error with two arguments.

raplima:
I think I can do that. I'm a little confused about what the function should return and `PyErr_Format` `PyErr_SetString` .

Does one of the options below work?
```
template< typename T1, typename T2 >
PyObject* RuntimeError( const char* msg, T1 t1, T2 t2 ) {
    return PyErr_Format( PyExc_RuntimeError, msg, t1, t2 );
}
```
or 
```
template< typename T1, typename T2 >
PyObject* RuntimeError( const char* msg, T1 t1, T2 t2 ) {
    PyErr_SetString( PyExc_RuntimeError, msg , t1, t2)
    return NULL;
}
```
I guess another quick option for me would be to remove the references to iline and xline number from the message for now.

jokva:
Usying PyErr_Format should be fine. Of course, removing the il/xl number from the message means you won't need the overload, which would also be ok (but maybe take something away from the patch).

jokva:
Alright, this looks good. I'll merge this now, but squash it to a single commit so that there are no non-compiling commits in the log.

Thanks for the contribution!

raplima:
Thanks @jokva! I squashed and pushed my branch. I also changed the name of this PR for something more appropriate (I didn't notice I could do that before). Let me know if I missed something

jokva:
It's already merged, don't worry :--)

PR #477, opened 2020-09-21
Return single sample in f.trace[i,j] for ints

jokva:
The intention was always for f.trace[i,j] to return a single sample when
i and j were both integers:

    assert f.trace[0, 1] == f.trace[0][1]

However, this evaluated to true when [i, j] returns an array, since the
result of the comparison is array([True]), and which is not falsy.

Reported in https://github.com/equinor/segyio/issues/475

### DIFF
--- a/python/segyio/trace.py
+++ b/python/segyio/trace.py
@@ -187,6 +187,7 @@ class Trace(Sequence):
             # trace to be loaded.
             j = slice(0, self.shape, 1)
 
+        single = False
         try:
             start, stop, step = j.indices(self.shape)
         except AttributeError:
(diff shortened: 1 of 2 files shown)


python/segyio/trace.py
```diff
@@ -201,7 +201,8 @@ def __getitem__(self, i):
         try:
             i = self.wrapindex(i)
             buf = np.zeros(n_elements, dtype = self.dtype)
-            return self.filehandle.gettr(buf, i, 1, 1, start, stop, step, n_elements)
+            tr = self.filehandle.gettr(buf, i, 1, 1, start, stop, step, n_elements)
+            return tr if n_elements > 1 else tr[0]
```
gsakkis, line 205: Does this also make slices of size 1 return a scalar? I.e. is `f.trace[5, 7:8]` identical to `f.trace[5, 7]`? I didn't test it but I think they should be different; the first one should return an array of size 1 (as it already does on master).

Issue #484, opened 2020-10-06
DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats


anders-kiaer:
When running with [`DepractionWarning`s turned on](https://docs.python.org/3/library/warnings.html#describing-warning-filters),
```
 WARNING:       [...]/python3.8/site-packages/segyio/line.py:75: DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  label: fread_trace0(label,
```
appears on Python3.8.

Looks to be related to https://github.com/python/cpython/commit/0895793194cd5f1153d2a286ed3a4fc876149b16

PR #490, opened 2020-10-14
Use Py_ssize_t for type-length arguments

jokva:
The with-length argument type specifier would be either int or
Py_ssize_t, depending on the PY_SSIZE_T_CLEAN macro being defined.

From the docs:
  For all # variants of formats (s#, y#, etc.), the type of the length
  argument (int or Py_ssize_t) is controlled by defining the macro
  PY_SSIZE_T_CLEAN before including Python.h. If the macro was defined,
  length is a Py_ssize_t rather than an int. This behavior will change
  in a future Python version to only support Py_ssize_t and drop int
  support.  It is best to always define PY_SSIZE_T_CLEAN.

https://docs.python.org/3/c-api/arg.html#arg-parsing

In python3.8, using ints for lengths (i.e. not defining
PY_SSIZE_T_CLEAN) is considered deprecated, and in future versions of
python only Py_ssize_t will be legal.

From the docs:
  It is recommended to always define PY_SSIZE_T_CLEAN before including
  Python.h. See Parsing arguments and building values for a description
  of this macro.

https://docs.python.org/3/c-api/intro.html

### DIFF
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -1,3 +1,4 @@
+#define PY_SSIZE_T_CLEAN
 #if defined(_DEBUG) && defined(_MSC_VER)
 #  define _CRT_NOFORCE_MAINFEST 1
 #  undef _DEBUG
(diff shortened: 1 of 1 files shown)


jokva:
Closes #484

gsakkis:
Could you please push a new PyPI release with this fix?

PR #500, opened 2021-01-04
Fix gather to sanitize the offsets slice

gsakkis:
Fixes https://github.com/equinor/segyio/issues/499


jokva:
This is a good change, thanks!

PR #524, opened 2022-02-03
Update wheel-building and version

ErlendHaa:


hakoneriksson (approved):
We should consider dropping support for older versions and go with manilinux2014 since 2010 uses CentOS 6 (EOL)
Perhaps next release should include musllinux as well, but this release should be fine.

PR #548, opened 2023-02-22
Apply DelayScalar to get correct samples offset

da-wad:
SEG-Y rev.1 gives the following info on trace header bytes 215-216:

> Scalar to be applied to times specified in Trace Header bytes 95-114 to give the true time value in milliseconds. 
Scalar = 1, +10, +100, +1000, or +10,000. 
If positive, scalar is used as a multiplier; 
if negative, scalar is used as divisor. 
A value of zero is assumed to be a scalar value of 1.

This scalar appeared not to be applied by segyio, making PR which should fix that.


python/segyio/open.py
```diff
@@ -172,8 +172,14 @@ def open(filename, mode="r", iline = 189,
     )
 
     try:
+        # Bytes 215-216
```
ErlendHaa, line 175: Noice? Maybe remove?

python/segyio/open.py
```diff
@@ -172,8 +172,14 @@ def open(filename, mode="r", iline = 189,
...
+        delay_scalar = f.header[0][segyio.TraceField.ScalarTraceHeader]
+        if delay_scalar == 0:  # A value of zero is assumed to be a scalar value of 1.
+            delay_scalar = 1
+        elif delay_scalar < 0:  # if negative, scalar is used as divisor.
```
ErlendHaa, line 179: Maybe remove these comment too. When I first read it I thought you made this assumption, but while checking the spec I see this is the intended behaviour. And the second one just explaining what's going on in the next line of code anyways

ErlendHaa (commented):
Nice catch! This is definitely what the spec says. Good contribution. Merge the 2 commits together remove some comments and we'll merge it

python/segyio/open.py
da-wad: Cleaner now 👍

dmlbarker:
I can confirm that this fixes segyio for a previously wrongly interpreted segy

ErlendHaa (approved):
LGTM

PR #563, opened 2023-11-20
Build python 3.12 wheels in Github Actions

sondreso:


python/pyproject.toml
```diff
@@ -61,5 +61,9 @@ manylinux-x86_64-image = "manylinux2010"
 select = "cp311*"
 manylinux-x86_64-image = "manylinux2014"
 
+[[tool.cibuildwheel.overrides]]
+select = "cp312*"
```
sveinung-r, line 65: I suggest including 3.12 in the override above like this: cp31{1,2}*

sveinung-r:
Not sure why the manylinux wheels for Python 3.9 is all of a sudden failing, but we can try building everything from 3.9 and above with manylinux2014. Or perhaps build everything with manylinux2014?

sondreso:
> [slice](https://docs.python.org/3/library/functions.html#slice) objects are now hashable, allowing them to be used as dict keys and set items. (Contributed by Will Bradshaw, Furkan Onder, and Raymond Hettinger in [gh-101264](https://github.com/python/cpython/issues/101264).)'

- From https://docs.python.org/3/whatsnew/3.12.html#other-language-changes

This seems to break some assumptions in the code from what I can tell. My knowledge of the inner workings of segyio is limited, so whoever wants to give it a try please feel free to hijack this PR and give it a spin :slightly_smiling_face:

sveinung-r (approved):

PR #570, opened 2023-12-08
Add include for cstdint, fix segyio build on fedora

SaidBenaissa:


ajaust:
Hi! Thanks for this PR. We have added some fixes in the most recent release. We did a quick test and segyio build fine on Fedora. Do you think that your PR is still needed?

ajaust:
I can confirm that the issue persist. I could reproduce it on Red Hat Linux 8. I will push some fixups and then try to get it merged soon.

lib/experimental/segyio/segyio.hpp
```diff
@@ -13,6 +13,7 @@
 #include <vector>
 
 #include <segyio/segy.h>
+#include <cstdint>
```
achaikou, line 16: This in theory should be near other common includes, not near segy.h one, but ok.

I even think other includes are sorted alphabetically, here and in the other file as well. But we can be evil and break sorting too here.

achaikou (approved):
Apply fixups/amends and let get this merged.

lib/experimental/segyio/segyio.hpp
ajaust: I just added another fixup to reorder it. I want it to be in order. 😄

achaikou (approved):
Looks nice! Rebase and merge.

yngve793 (approved):
I compiled and built it on Redhat 8.1. For some reason it still builds and passes tests with and without  "#include <cstdint>".

No other comments.

lib/CMakeLists.txt
```diff
@@ -138,3 +138,19 @@ add_test(NAME c.segy.mmap     COMMAND c.segy --test-mmap)
...
+# above because the test framework does include some C++ dependencies like
+# `cstdint` and `algorithm` that may be used in the segyio headers.
+add_executable(c-include.segy test/test-include.c)
+target_link_libraries(c-include.segy segyio)
+target_include_directories(c-include.segy PRIVATE src)
+add_test(NAME c-include.segy COMMAND c-include.segy)
+
+add_executable(cpp-include.segy test/test-include.cpp)
+target_link_libraries(cpp-include.segy segyio::segyio)
+target_include_directories(cpp-include.segy PRIVATE src experimental)
```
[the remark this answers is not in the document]
ajaust: Good catch.

ajaust:
That the tests sometimes pass without includes seems to be due to the used compilers and their standard libraries.

Newer versions of GCC (I tested 13) seem to complain about missing `cstdint` and `algorithm`. Microsoft compiler complains also about `stdexcept` missing. Clang on our Apple machines (currently Apple Clang 17) did not complain much.

PR #576, opened 2024-07-26
Fix attribute error in setup.py

sveinung-r:
This line was causing a  error due to an update to scikit-build.  The issue was that the setuptools.command.test module is not put into the symbol table by the setuptools import, but it was put there during the skbuild import causing it to be available. Due to changes in scikit-build this is no longer the case and the line gives an AttributError.

The rationale for this line was that scikit-builds test command implied develop (this was obnoxious), something that is no longer true. There is thus no longer any reason to keep this line, so we can fix this issue by simply removing it.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,7 +23,7 @@ jobs:
             analyzers: "cppcheck"
             scan: "scan-build --status-bugs"
             mkdoc: "-DBUILD_DOC=ON -DSPHINX_ARGS=-WT"
-          - os: macos-13-xlarge
+          - os: macos-latest
             privledges: "sudo"
             arch: arm64
 
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -25,7 +25,7 @@ jobs:
             arch: i686
           - os: macos-latest
             arch: x86_64
-          - os: macos-13-xlarge
+          - os: macos-latest
             arch: arm64
 
     steps:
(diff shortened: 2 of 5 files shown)


tjni:
@sveinung-r Hello! I'm trying to fix the build for this package in [nixpkgs](https://github.com/NixOS/nixpkgs), which is broken due to this error.

1. I notice quite a few commits. Are they unrelated to fix CI issues, or are they related to this change?
2. What's the status of this patch being merged?

Thank you for creating this patch and doing the leg work of figuring out what is going on.

ajaust:
Hei @tjni,

Awesome that you package `segyio` for Nix. :) 

I cannot comment on the first point, but I think it is still the plan to merge the PR.
The PR was blocked due to the failing test on aarch64. The issue was tracked down to be an issue with [`np.unique`](https://github.com/numpy/numpy/issues/27273). Afaik that was fixed, but we were still waiting for a new release of NumPy for an older Python release that we support (I think 3.9). I see that there are several new NumPy releases. We will check if can merge this PR soon.

ajaust (commented):
I had a look at the changes and they look good to me. 🎉 

I am uncertain if the [bug you reported](https://github.com/numpy/numpy/issues/27273) is fixed for Python 3.9. The [PR that fixes the reported bug](https://github.com/numpy/numpy/pull/27438) is part of NumPy 2.1.2, see [NumPy releases](https://github.com/numpy/numpy/releases), for which there is no Python 3.9 release. At least all of our tests passed the last time, but I am not sure if this was a lucky coincidence. Do you have any thoughts on this?

ajaust (approved):
Looks good. We will merge. The CI issues will be investigated/fixed in a separate PR as some of these issues have appeared recently and are out of scope of this PR.

PR #591, opened 2025-01-31
Pin versions of documenation dependencies

ajaust:
The versions are pinned to versions suggested by Snyk. The pinning was done manually because Snyk was not able to create a PR itself.


ajaust:
I will close this. Snyk showed several issues in dependencies because the repository was scanned using Python 3.7.

PR #594, opened 2025-02-03
Bring back old readthedocs look

achaikou:
Looks like after fixing readthedocs builds we need to specify the theme explicitly.

Also sphinx got updated. In new version content tree doesn't look the same anymore and is cluttered with domain objects. This new feature needs to be turned off to look the same as before.

Currently:

Before:

With this PR:

### DIFF
--- a/python/docs/conf.py
+++ b/python/docs/conf.py
@@ -50,6 +50,7 @@ extensions = [
     'sphinx.ext.intersphinx',
     'sphinx.ext.imgmath',
     'sphinx_gallery.gen_gallery',
+    'sphinx_rtd_theme',
 ]
 
 autosummary_generate = True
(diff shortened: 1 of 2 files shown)


yngve793 (changes requested):
The docs looks much better. :-)

Documentation says Copyright 2018. We should probably update. 

I need to update add the following packages to requirements-doc.txt:
- setuptools
- sphinx_rtd_theme

For the file requirements-readthedocs.txt 
"pip install -r requirements-readthedocs.txt" fails with 
ModuleNotFoundError: No module named 'skbuild'

What is the difference between requirements-doc.txt and requirements-readthedocs.txt.

yngve793 (approved):
Comments checked out.

PR #597, opened 2025-02-13
Add benchmark tests

achaikou:
Tests are not run by default, need to specifically run `pytest test/benchmarks.py` to trigger.

I reduced file size even more, so it is now around 3GB. Not sure anymore it helped though.

Out of last 10 attempts 1 [already flopped](https://github.com/achaikou/segyio/actions/runs/13310423399/job/37172257016). For the whole run all tests were1.25x- 2x times slower than usual.

So it probably mostly depends on current workload of the machine we are getting. Figuring out how to get dedicated machine just for this is too much, I think.

So maybe we want to remove running this for pull request / push to master?
Automatically re-running the job in case of failure isn't straightforward, so we'll have to retry manually if something is off.

### DIFF
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,0 +1,44 @@
+name: Benchmarks
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+  workflow_dispatch:
+    inputs:
+      commit_ref:
+        description: "optional: compare to specific ref"
+        required: false
+
+jobs:
+  benchmarks:
+    name: Benchmark scripts
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install build/test dependencies
+        shell: bash
+        working-directory: python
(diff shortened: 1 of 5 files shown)


.github/workflows/benchmarks.yaml
```diff
@@ -0,0 +1,44 @@
...
+        required: false
+
+jobs:
+  benchmarks:
+    name: Benchmark scripts
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+
```
ajaust, line 21: Do you think we would gain anything (certainty about Python version, runtime stability) if we would explicitly select a Python version by using https://github.com/actions/setup-python?

ajaust (commented):
This looks good to me overall. I left some comments for clarification, but I don't think there is anything major.

I would suggest to run the benchmarks when pushing to master and in PRs to master, i.e., leave it as it is. Then let's observe how often it fails and how much it annoys us.

ajaust (approved):
Looks good. 🎉

PR #600, opened 2025-02-25
Fix declaration without a prototype warning

yngve793:
During compiling the functions without parameter definition gives warnings.


achaikou (approved):
Looks like it is a proper solution for the problem!
I will add clang Debug job for the future so that it gets printed on CI as well.

PR #601, opened 2025-02-25
Use 64 bits to read and write field

yngve793:
To prepare for the new 8-bytes fields in the binary header the field get and set methods are updated to use int64_t.

### DIFF
--- a/applications/apputils.c
+++ b/applications/apputils.c
@@ -36,7 +36,7 @@ int parseint( const char* str, int* x ) {
 }
 
 int bfield( const char* header, int field ) {
-    int32_t f;
+    int64_t f;
     int err = segy_get_bfield( header, field, &f );
 
     if( err ) return -1;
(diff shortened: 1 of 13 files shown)


achaikou (commented):
Huh, this is a difficult one. :thinking: 
[Probably would also benefit from cppcheck-debugging CI in place - something that would catch all narrowing implicit conversions.]

Not sure if merging this change without dealing with context at the same time is a good idea. I don't see a way how to mark all of that with "should be tracked later".
Only way is we promise that after everything is done, every single function would be looked at and its logic scrutinized?

What is your vision on dealing with this?

---
We also would need to decide between exchanging whole hierarchy with 64-ints or actually checking specification beforehand and casting short ints to 32.
It seems like many of important values can be 64 ints now, so maybe full call hierarchy updat makes more sense?...

yngve793:
No longer relevant.

PR #603, opened 2025-02-27
Add more checks to build workflows

achaikou:
Hopefully these workflows will help us with coming updates, now that we know that we will be drowning in type changes, and will help us focus on making conscious type choices.

Main goal was to have a clean warnings state. Some fixes are forced conversion because of mismatched current types. Some are not fixed, just ignored, as I have no clue how to do that.

### DIFF
--- a/.github/workflows/analyzers.yaml
+++ b/.github/workflows/analyzers.yaml
@@ -1,0 +1,98 @@
+name: Run analyzers
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch:
+
+jobs:
+  cppcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install cppcheck
+
+      - name: Configure
+        run: |
+          cmake -S . -B build                     \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON    \
+            -DBUILD_PYTHON=OFF                    \
+            -DBUILD_TESTING=OFF
+
+      - name: Run cppcheck
+        run: |
+          cppcheck                                          \
+            --enable=style,portability,performance,warning  \
+            --library=posix                                 \
+            --library=cppcheck/segyio.cfg                   \
+            --suppressions-list=cppcheck/suppressions.txt   \
+            --inline-suppr                                  \
+            --project=build/compile_commands.json           \
+            --error-exitcode=1
+
+  scan_build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install clang clang-tools libfindbin-libs-perl
+
+      - name: Configure segyio
+        run: |
+          scan-build --status-bugs   \
+            cmake                    \
+            -S .                     \
+            -B build                 \
+            -DCMAKE_BUILD_TYPE=Debug \
+            -DBUILD_PYTHON=OFF       \
+            -DBUILD_TESTING=OFF
+
+      - name: Run scan-build
+        run: |
+          scan-build --status-bugs \
+            cmake                  \
+            --build build
+
+  valgrind:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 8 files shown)


.github/workflows/analyzers.yaml
```diff
@@ -0,0 +1,95 @@
...
+jobs:
+  cppcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install cppcheck
```
ajaust, line 19: Do you have any opinion on installing cppcheck via Ubuntu's package manager vs. compiling it from source to get a more recent version? In `si4ti` cppcheck is/was compiled from source and I am wondering if either way is preferable over the other.

.github/workflows/analyzers.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install valgrind
+
+      - name: Build segyio
+        run: |
+          cmake -S . -B build           \
+            -DCMAKE_BUILD_TYPE=Release  \
```
ajaust, line 77:
```suggestion
            -DCMAKE_BUILD_TYPE=Debug  \
```

I think running Valgrind on a debug build gives fewer false positives and better information (stack traces) if something goes wrong.

From [Valgrind's quickstart guide (point 2)](https://valgrind.org/docs/manual/quick-start.html):

> Compile your program with -g to include debugging information so that Memcheck's error messages include exact line numbers. Using -O0 is also a good idea, if you can tolerate the slowdown. With -O1 line numbers in error messages can be inaccurate, although generally speaking running Memcheck on code compiled at -O1 works fairly well, and the speed improvement compared to running -O0 is quite significant. Use of -O2 and above is not recommended as Memcheck occasionally reports uninitialised-value errors which don't really exist.

.github/workflows/analyzers.yaml
```diff
@@ -0,0 +1,95 @@
...
+        run: |
+          sudo apt-get update
+          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install clang clang-tools libfindbin-libs-perl
+
+      - name: Configure segyio
+        run: |
+          sudo scan-build --status-bugs cmake \
+            -S .                              \
+            -B build                          \
+            -DCMAKE_BUILD_TYPE=Release        \
```
ajaust, line 54:
```suggestion
            -DCMAKE_BUILD_TYPE=Debug          \
```
I have no experience with scan-build, but it the [documentation recommends to use a debug build](https://clang.llvm.org/docs/analyzer/user-docs/CommandLineUsage.html#always-analyze-a-project-in-its-debug-configuration).

.github/workflows/build.yaml
```diff
@@ -47,8 +69,11 @@ jobs:
             -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
             -DBUILD_PYTHON=OFF \
             -DBUILD_SHARED_LIBS=ON \
-            -DCMAKE_BUILD_TYPE=Release \
+            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+            -DCMAKE_COMPILE_WARNING_AS_ERROR=ON \
```
ajaust, line 73: I am impressed that the tests actually pass with such strict checks.

.github/workflows/build.yaml
```diff
@@ -10,24 +10,46 @@ on:
...
             cmake_generator: "-G \"Visual Studio 16 2019\" -A Win32"
+            build_type: "Release"
+            compiler_flags: "-D_CRT_SECURE_NO_WARNINGS /EHsc"
           - os: ubuntu-latest
             privledges: "sudo"
+            build_type: "Release"
+          - os: ubuntu-latest
+            privledges: "sudo"
+            build_type: "Debug"
+            compiler_flags: "-Wextra -Wall"
```
ajaust, line 36: We could add [`-pedantic`](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html#Options-to-Request-or-Suppress-Warnings) if we care about ISO C++ compliance

ajaust (commented):
Looks good overall. 👍 I left some minor comments w.r.t. to compiler flags and build targets. I hope these comments make sense. 😺

ajaust (approved):
Looks good. Feel free to squash and merge. 🙂

PR #606, opened 2025-03-25
Suppress cppcheck exceptions

yngve793:
The objective is to run cppcheck during rebase testing. Adding suppress comments such that the codebase can pass without error or warning flags being raised.

### DIFF
--- a/.github/workflows/analyzers.yaml
+++ b/.github/workflows/analyzers.yaml
@@ -9,14 +9,22 @@ on:
 
 jobs:
   cppcheck:
-    runs-on: ubuntu-latest
+    name: cppcheck on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            cppcheck_install_command: sudo apt-get update && sudo DEBIAN_FRONTEND=noninteractive apt-get -y install cppcheck
+          - os: macos-latest
+            cppcheck_install_command: brew install cppcheck
     steps:
       - uses: actions/checkout@v4
 
       - name: Install dependencies
         run: |
-          sudo apt-get update
-          sudo DEBIAN_FRONTEND=noninteractive apt-get -y install cppcheck
+          ${{ matrix.cppcheck_install_command }}
 
       - name: Configure
         run: |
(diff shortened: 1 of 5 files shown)


lib/test/segy.cpp
```diff
@@ -1762,7 +1762,7 @@ SCENARIO( "reading a 2-byte int file", "[c.segy][2-byte]" ) {
             CHECK( trace_bsize == 75 * 2 );
         }
 
-        WHEN( "the format is valid" ) THEN( "setting format succeeds" ) {
+        WHEN( "the format is valid" ) THEN( "setting format succeeds" ) { // cppcheck-suppress unknownMacro
```
achaikou, line 1765: When and Then are not unknown macros. Maybe the format of this test could be just changed to
```
WHEN("") {
    THEN("") {
  }
}
```
?
It seems weird to suppress this instead of fixing it.

Also shouldn't line below (1770) fail with the same?

yngve793: It does not seams to work. At least I get the same error with brackets.

yngve793 (commented):
Added a commit with job that runs analyzers on mac. It is not clear how to fix the issues.

yngve793 (commented):
Updated analyzers workflow as suggested. 
Added code suggestions to remove mac cpp-checks.

.github/workflows/analyzers.yaml
```diff
@@ -34,7 +34,36 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          brew install cppcheck
+
+      - name: Configure
+        run: |
+          cmake -S . -B build                     \
+            -DCMAKE_EXPORT_COMPILE_COMMANDS=ON    \
+            -DBUILD_PYTHON=OFF                    \
+            -DBUILD_TESTING=OFF
```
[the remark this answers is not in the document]
achaikou: So do you want to enable cppcheck for tests here or not?

I am fine with both, it is just if you want to use them locally and we won't have them on CI, you might get errors after someone else merges their PR.

yngve793:
Not relevant any more.

achaikou (approved):
As we agreed to not turn on cppcheck on tests, then I am happy with this PR - rebase and merge at will :cat:

PR #609, opened 2025-03-27
Fix benchmark CI

achaikou:
Current logic is flawed.
Intention of benchmark tests was to build library from two different commits. However only C/C++ part was built separately, python code was not touched.

When current commit's directory was checked out to run the fresh tests, so was other python code: tests were always run on the same python part of the library.

To fix that, build a local wheel from existing code and make sure it is being used instead of local code.

### DIFF
--- a/.github/actions/benchmark/action.yaml
+++ b/.github/actions/benchmark/action.yaml
@@ -18,31 +18,35 @@ runs:
     - name: Checkout commit which performance should be measured
       shell: bash
       run: |
+        git restore .
         git fetch origin ${{ inputs.ref }} --depth 1
         git checkout ${{ inputs.ref }}
 
-    - name: Configure
+    - name: Build wheel
+      shell: bash
+      env:
+        CIBW_BUILD_VERBOSITY: 1
+      run: |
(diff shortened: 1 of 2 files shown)


.github/actions/benchmark/action.yaml
```diff
@@ -18,32 +18,31 @@ runs:
...
       run: |
+        git restore .
         git fetch origin ${{ inputs.ref }} --depth 1
         git checkout ${{ inputs.ref }}
 
-    - name: Configure
+    - name: Build wheel
       shell: bash
       run: |
-        cmake -S . -B build \
```
yngve793, line 27: Is the cmake config and build included in the pyhton wheel build?

.github/workflows/benchmarks.yaml
```diff
@@ -19,8 +19,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Install wheel and cibuildwheel
+        run: python3 -m pip install wheel cibuildwheel
+
       - name: Install build/test dependencies
-        shell: bash
```
yngve793, line 23: Out of curiosity. Why remove bash?

yngve793 (approved):
Not sure how to test that the two wheels are different. 
Looks good and all tests and checks passes.

PR #610, opened 2025-04-03
Bump minimum CMake version to 3.11

ajaust:
Red Hat Enterprise Linux (RHEL) 7 has reached its end of life in 2024. This means we can bump the minimum required version to 3.11 which is provided in RHEL 8 [1]. We assume that other systems either have CMake 3.11 or newer available, or have a way to install a sufficently new CMake versions.

Bumping the minimum version removes some deprecation warning (about support for older CMake versions being dropped), and also ensures compatibility with CMake 4.0 which has dropped support for CMake <3.5 and was rolled out to GitHub runners recently.

[1]: https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html-single/8.0_release_notes/index#platform-tools

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.11)
 project(segyio LANGUAGES C CXX)
 
 include(CheckFunctionExists)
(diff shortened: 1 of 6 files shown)


achaikou (commented):
Awesome.
1. Should we also update cmake version in setup-CMakeLists.txt 
2. Should the version in README be updated?

achaikou (approved):
Looks good to me!
Rebase and merge.

Minimum required version makes sense as Red Hat 8 seems to have oldest cmake from currently alive major systems.
Also segyio 2.0 is a good opportunity to bump required cmake several versions higher.

PR #612, opened 2025-04-15
Bigendian speedup

achaikou:
Assumption that bigendian tests could be slow didn't hold for long, so they are sped up to 10x.


yngve793 (approved):
Yes, this was faster and the cache worked.

PR #613, opened 2025-04-16
Refactoring preparations

achaikou:
A few small changes that is convenient to have before C datasource introduction.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -185,6 +185,26 @@ static inline void native_ibm( void* buf ) {
     memcpy( buf, &u, sizeof( u ) );
 }
 
+static void bswap64_mem( char* dst, const char* src ) {
+    uint64_t v;
+    memcpy( &v, src, 8 );
+    v = bswap64( v );
+    memcpy( dst, &v, 8 );
+}
+static void bswap32_mem( char* dst, const char* src ) {
+    uint32_t v;
+    memcpy( &v, src, 4 );
+    v = bswap32( v );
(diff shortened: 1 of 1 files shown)


lib/src/segy.c
```diff
@@ -2016,20 +1995,31 @@ int segy_writesubtr( segy_file* fp,
 
     step *= elemsize;
 
+    void ( *bswap_mem )( char*, const char* );
+    if ( elemsize == 8 ) {
```
yngve793, line 1999: The if check is still inside the loop. I do not understand what we are trying to achieve here. Original code check datatype, scanned through a buffer, read out a value into a variable do a bswapXX and write it back into a new buffer. This code is doing exactly the same but the checks for each element is inside the loop. If the goal is to speed things up we could copy the hole buffer in one go, check the datatype and scan through with a pointer and preform bswapXX without copying every single value to an external variable. This will save a lot of read and write operations.

yngve793 (commented):
The objective for changing the  bswapXX loop is still unclear. The original is faster do to the internal if check in the new version.

lib/src/segy.c
achaikou: Where do you see the loop here?? :scream_cat: 

All this change now does is _restructure_ the code, no logic is changed, only function pointers are introduced.

yngve793: Ok, I see that you use some pointer trick outside the loop. Is this ok if casting is bad?

Why are we doing this and what are we gaining from it?

yngve793 (approved):

PR #616, opened 2025-04-28
Fix of recently failing CI

achaikou:
- Today's benchmark test fixes
- Fixes to the binary header values

Note: I suspect that file creation through `tools.from_array` might fail for users as set trace fields are 2 bytes (trace number, sample count). We just don't have big enough tests to confirm this. Probably would be best to create them one we have extended trace headers in place.


yngve793 (approved):
Ok

PR #617, opened 2025-05-02
Use fseeko and _fseeki64 for file seek

achaikou:
I am uncertain about this one.
Main question is: why hasn't it been done from the start?
Also: could this change actually make code stop working for some people? Unless they are running segyio on a washing machine, I doubt so, but still.

Code is updated to be similar to "tell" part.


lib/src/segy.c
```diff
@@ -562,26 +562,12 @@ static int segy_seek( segy_file* fp,
 #endif //HAVE_MMAP
 
     int err;
-#if LONG_MAX == LLONG_MAX
-    assert( pos <= LONG_MAX );
-    err = fseek( fp->fp, (long)pos, SEEK_SET );
+#ifdef HAVE_FSEEKO
+    err = fseeko( fp->fp, pos, SEEK_SET );
```
ajaust, line 566: Do we need to ensure to always use a 64-bit integer or does it not matter? 

From the Microsoft documentation, I understand that `_fseeki64` will always use a 64-bit integer to represent the offset.

The Linux (POSIX) [definition](https://pubs.opengroup.org/onlinepubs/9699919799/) seems to be vague what `off_t` is. From some [discussions on Stackoverflow](https://stackoverflow.com/a/1035750) it appears that one might have to enforce 64-bit offset types by a compile definition. I assume we are happy with the POSIX definition of `fseeko` because it would just use and cast to the correct type itself (?). Is this understanding correct?

CMakeLists.txt
```diff
@@ -87,6 +87,10 @@ if (HAVE_SYS_STAT_H)
     if (HAVE_FTELLI64)
         list(APPEND fstat -DHAVE_FTELLI64)
     endif ()
+    check_function_exists(_fseeki64 HAVE_FSEEKI64)
+    if (HAVE_FSEEKI64)
+        list(APPEND fstat -DHAVE_FSEEKI64)
```
ajaust, line 92: Do you know why the Windows (?) methods (`ftelli64` etc.) ending on `i64` are added to the `fstat` list, but the POSIX methods ending on `o` (`ftello` etc.) are getting their own individual lists?

ajaust (commented):
Replacing custom code by already available standardised methods is a good idea. 👍 

I just left some comments to understand better what is going on. I have not used `fseek` and its variants myself yet.

achaikou (commented):
Very good points, thank you! :blue_heart:   Comments are replied to, but nothing is changed yet, because I remembered today we solve the same issue in lfp.

[There](
https://github.com/equinor/layered-file-protocols/blob/master/src/cfile.cpp#L19) we use `fseeko`/`_fseeki64` only when long long is the same as long.

Don't look at me, I have no memory of why it was done the way it is :smile:

So I do not 100% see reason behind it, but maybe we should follow in lfp footsteps and try implementing it somewhat similar?
As we are in C, not C++, enable-if templates won't work, but original segyio `#if LONG_MAX == LLONG_MAX` check should still be enough.

lib/src/segy.c
```diff
@@ -562,26 +562,12 @@ static int segy_seek( segy_file* fp,
...
-    while( pos >= LONG_MAX && err == SEGY_OK ) {
-        err = fseek( fp->fp, LONG_MAX, SEEK_CUR );
-        pos -= LONG_MAX;
-    }
-
-    if( err != 0 ) return SEGY_FSEEK_ERROR;
-
-    assert( pos <= LONG_MAX );
-    err = fseek( fp->fp, (long)pos, SEEK_CUR );
+    assert( false );
```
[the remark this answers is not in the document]
achaikou: :dizzy_face: 
Wait. So apparently static_assert is difficult.
We do not specify any C standard in cmake file, but it is likely less than C23.

So if we are on compiler before C11 (could we be?), we have nothing according to standard.
If we are on a reasonably new compiler, then it is `_Static_assert`.
If we are on windows, it is apparently `static_assert` , because `unresolved external symbol _Static_assert` ...

But if we are on an unknown system that has no `_fseeki64` or `fseeko` or anything, what should we call then?? Anything at all because we just want it to fail?

achaikou (commented):
Thanks for the review!
#error is implemented and _FILE_OFFSET_BITS was already set to 64.

ajaust (approved):
Looks good to me. Squash the fixups and then merge. 👍

PR #619, opened 2025-05-08
Remove bfield and trfield

yngve793:
In preparation for reading fields with data type float I want to simplify the code. 

Functions bfield and trfield returns -1 in case of error. For most fields -1 is a valid value. Not all cases using these functions check for -1. Remove the functions and ignore the error code in cases it is not checked.

### DIFF
--- a/applications/apputils.c
+++ b/applications/apputils.c
@@ -35,22 +35,6 @@ int parseint( const char* str, int* x ) {
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


achaikou (commented):
Agree with you comment.
Otherwise all looks nice :+1:

achaikou (approved):

PR #620, opened 2025-05-09
Impr/add binary fields

yngve793:
Add binary fields defined in SEGY-Y 2.1 and 2.0.
Introduce code to read and write uint64 and float64 fields.


lib/src/segy.c
```diff
@@ -692,6 +737,7 @@ static int fd_get_int( const segy_field_data* fd, int* val ) {
 }
 
 static int init_segy_field_data(int field, segy_field_data* fd) {
+    fd->value.u64 = 0;
```
yngve793, line 740: Writing a float64 and reading the union buffer as u64 is not trivial to do without one or more of the tests failing. This line is added as a workaround to let the tests pass.

lib/src/segy.c
```diff
@@ -801,6 +881,11 @@ static int set_field( char* header,
             memcpy( header + (fd->field_index - 1), &(fv.u8), formatsize( fd->datatype ));
             return SEGY_OK;
 
+        case SEGY_IEEE_FLOAT_8_BYTE:
+            fv.u64 = htobe64( fv.u64 );
```
achaikou, line 885: Not sure if it is related to `fd->value.u64 = 0;` discussion.

I wonder whether floating point numbers should have their own `htobe`/`betoh`. 

New version of endiannes library (the one we are using in dlisio and maybe would one day want to switch in segyio too???) does c-conversion with the help of union
https://gist.github.com/jtbr/7a43e6281e6cca353b33ee501421860c#file-endianness-h-L146

Though if you do not experience any issues with using uint htobe for float, then there is probably no need to do anything about it, at least yet.

achaikou (commented):
Good to see that we are getting to the fun parts. :sunglasses: 
Bad that we might be pioneers and still don't have real various 2.1 files... :unamused:

yngve793 (commented):
New suggestions added.

yngve793 (commented):
Updated

achaikou (approved):
Rebase and merge! :tada:

achaikou (approved):

PR #621, opened 2025-05-09
Update application segyio-catb to use the full range of binary header fields

yngve793:
Getting access to both filed value and field datatype the code in segyio-catb can be simplified. 
Test outputs are updated to contain the new fields.


applications/segyio-catb.c
```diff
@@ -210,20 +149,41 @@ int main( int argc, char** argv ){
...
+
+            if( opts.nonzero && fd.value.u32 == 0) continue;
+
+            int byte_offset = (field_data[c].offset - SEGY_TEXT_HEADER_SIZE);
+            const char* short_name = field_data[c].short_name;
+            const char* description = field_data[c].description;
+
+            if( fd.datatype == SEGY_UNSIGNED_INTEGER_8_BYTE ) {
+                long long int field = fd.value.u64;
+                if( opts.description )
```
achaikou, line 165: Preference note, disregard it if you want :cat: :
I often get burned by two-lines ifs and fors when they don't have {}. I overlook them not having {}, add some debug statement "inside" and then for the next 20 minutes wonder WTF is going on... :pouting_cat: 

So I would really appreciate adding braces to this if/else :sun_with_face:

achaikou (commented):
Nice, that was easy to follow through :+1: 

Only last 4 commits are looked at, assuming first 4 are just copies.

Note to commit message:
Add fields -> Add new fields to segyio-catb.c ?
"Add fields sound too generic".

yngve793 (commented):
Thanks for pointing out the improvements in the short name and descriptive fields of segyio-catb. If anybody have a clear idea of how to name things and define descriptions this can be discussed.

applications/segyio-catb.c
```diff
@@ -113,9 +113,25 @@ int main( int argc, char** argv ){
...
         {SEGY_BIN_SEGY_REVISION, "rev", "SEG Y Format Revision Number"},
+        {SEGY_BIN_SEGY_REVISION_MINOR, "revmin", "SEG Y Format Min Revision Number"},
         {SEGY_BIN_TRACE_FLAG, "trflag", "Fixed length trace flag"},
         {SEGY_BIN_EXT_HEADERS, "exth", "Number of 3200-byte, Extended Textual File Headers"},
+        {SEGY_BIN_MAX_ADD_TRACE_HEADERS, "maxtrh", "MaxTrh"},
+        {SEGY_BIN_SURVEY_TYPE, "survty", "SurveyType"},
+        {SEGY_BIN_TIME_BASIC_CODE, "timbas", "TimeBasis"},
+        {SEGY_BIN_NR_TRACES_IN_STREAM, "ntrpr", "NrTracesInStream"},
+        {SEGY_BIN_FIRST_TRACE_OFFSET, "lcf", "FirstTraceOffset"},
+        {SEGY_BIN_NR_TRAILER_RECORDS, "ntr", "NrTrailerRecords"}
```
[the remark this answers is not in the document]
yngve793: Ok

yngve793 (commented):
Description updated.

achaikou (commented):
Looks good to me. Prepare for merge :) :cat: 
I'll approve when it is rebased over the other PR because it is a bit difficult to see the full picture with those duplicated commits.

applications/segyio-catb.c
```diff
@@ -210,20 +149,41 @@ int main( int argc, char** argv ){
...
-                    su[ c ],
-                    field,
-                    byte_offset,
-                    su_desc[ c ] );
+        int nr_fields = sizeof(field_data)/sizeof(binary_field_type);
+        for( int c = 0; c < nr_fields; ++c ){
+            segy_field_data fd = segy_get_field( binheader, field_data[c].offset );
+            if ( fd.error ) return errmsg( fd.error, "Unable to read field" );
+
+            if( opts.nonzero && fd.value.u32 == 0) continue;
```
achaikou, line 157: Oh, so this is the problem... Honestly missed this line when reading.

Maybe it has something to do with environments and compilers, after you removed u64=0 assignment?
Anyway it would look better inside datatype "if"s, so good that CI noticed that.

PR #622, opened 2025-05-14
BYOIO: changes in C core

achaikou:
First part of changes allowing users to supply their own IO.
This one makes changes only to internal library, opening up for introducing new functionality in python later.

Note: I can't figure out if we shouldn't just delete cppcheck segyio.cfg - I am not sure if it helps cppcheck to know about possible leaks (then it should stay and be maintained) or if it is just more or less useless for us.
This is a question for future. For now I just updated there function signatures.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -411,6 +411,173 @@ static int file_size( FILE* fp, long long* size ) {
 }
 #endif //HAVE_SYS_STAT_H
 
+static int fileread( segy_datasource* self, void* buffer, size_t size ) {
+    FILE* file = (FILE*)self->stream;
+    size_t readc = fread( buffer, size, 1, file );
+    if ( readc != 1 ) return SEGY_FREAD_ERROR;
+    return SEGY_OK;
+}
+
+static int filewrite( segy_datasource* self, const void* buffer, size_t size ) {
+    if ( !self->writable ) return SEGY_READONLY;
+    FILE* file = (FILE*)self->stream;
+    size_t writec = fwrite( buffer, size, 1, file );
+    if ( writec != 1 ) return SEGY_FWRITE_ERROR;
+    return SEGY_OK;
+}
+
+static int fileseek( segy_datasource* self, long long pos, int whence ) {
+    FILE* file = (FILE*)self->stream;
+    int err;
+#if LONG_MAX == LLONG_MAX
+    err = fseek( file, (long)pos, whence );
+#elif HAVE_FSEEKO
+    err = fseeko( file, pos, whence );
+#elif HAVE_FSEEKI64
+    err = _fseeki64( file, pos, whence );
+#else
+    error "no 64-bit alternative to fseek() found, and long is too small"
+#endif
+    if( err != 0 ) return SEGY_FSEEK_ERROR;
+    return SEGY_OK;
+}
+
+static int filetell( segy_datasource* self, long long* pos ) {
+    FILE* file = (FILE*)self->stream;
+#if LONG_MAX == LLONG_MAX
+    *pos = ftell( file );
+#elif HAVE_FTELLO
+    *pos = ftello( file );
+#elif HAVE_FTELLI64
+    *pos = _ftelli64( file );
+#else
+    error "no 64-bit alternative to ftell() found, and long is too small"
+#endif
+    return SEGY_OK;
+}
+
+static int filesize( segy_datasource* self, long long* out ) {
+    FILE* file = (FILE*)self->stream;
+    int err = file_size( file, out );
+    return err;
+}
+
+static int fileflush( segy_datasource* self ) {
+    // flush is a no-op for read-only files
+    if ( !self->writable ) return SEGY_OK;
+    FILE* file = (FILE*)self->stream;
+
+    int flusherr = fflush( file );
+    if ( flusherr != 0 ) return SEGY_FWRITE_ERROR;
+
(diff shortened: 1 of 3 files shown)


yngve793 (commented):
Generally it looks good. 
Suggest to split the last commit into smaller parts. Style issues and introduction of new functions can be separated out into new commits without affecting tests.

lib/src/segy.c
```diff
@@ -2262,66 +2128,73 @@ int segy_readsubtr( segy_file* fp,
...
+            err = ds->seek( ds, step - elemsize, SEEK_CUR );
+            if( err != 0 ) return SEGY_DS_SEEK_ERROR;
+        }
 
-        if (fp->lsb) {
-            if (fp->elemsize == 8) bswap64vec(buf, slicelen);
-            if (fp->elemsize == 4) bswap32vec(buf, slicelen);
-            if (fp->elemsize == 3) bswap24vec(buf, slicelen);
-            if (fp->elemsize == 2) bswap16vec(buf, slicelen);
+        if( ds->lsb ) {
```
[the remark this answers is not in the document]
yngve793: Suggested change for style commit:

diff --git a/lib/src/segy.c b/lib/src/segy.c
index b2d1fcc..2cedcc6 100644
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -2400,11 +2400,11 @@ int segy_writesubtr( segy_file* fp,
     step *= elemsize;
 
     void ( *bswap_mem )( char*, const char* );
-    if ( elemsize == 8 ) {
+    if( elemsize == 8 ) {
         bswap_mem = bswap64_mem;
-    } else if ( elemsize == 4 ) {
+    } else if( elemsize == 4 ) {
         bswap_mem = bswap32_mem;
-    } else if ( elemsize == 2 ) {
+    } else if( elemsize == 2 ) {
         bswap_mem = bswap16_mem;
     } else {
         return SEGY_INVALID_ARGS;
@@ -2445,12 +2445,12 @@ int segy_writesubtr( segy_file* fp,
 
     char* cur = (char*)tracebuf + elemsize * defstart;
     if ( !fp->lsb ) {
-        for ( ; slicelen > 0; cur += step, --slicelen, src += elemsize ) {
-            memcpy( cur, src, elemsize );
+        for (; slicelen > 0; cur += step, --slicelen, src += elemsize) {
+            memcpy(cur, src, elemsize);
         }
     } else {
-        for ( ; slicelen > 0; cur += step, --slicelen, src += elemsize ) {
-            bswap_mem( cur, src );
+        for (; slicelen > 0; cur += step, --slicelen, src += elemsize) {
+            bswap_mem(cur, src);
         }
     }

achaikou (commented):
Thanks for the review! :cat:  Style commit is force-pushed before the other ones.

yngve793 (approved):
Apart from a merge conflict when rebasing it looks good.

lib/src/segy.c
```diff
@@ -2126,15 +2126,24 @@ int segy_readsubtr( segy_datasource* ds,
     char* dst = (char*)buf;
 
     if( !ds->minimize_requests_number ) {
-        err = ds->seek( ds, elemsize * defstart, SEEK_CUR );
-        if( err != 0 ) return SEGY_DS_SEEK_ERROR;
+        if( ds->read == memread && ds->seek == memseek ) {
```
yngve793: I would suggest to put a flag in in datasource instead of doing this slightly cryptic test. It is quite hard for future users to deduct that `if( ds->read == memread && ds->seek == memseek )` should imply memory.

yngve793 (commented):
Suggest to add info in datasource.

lib/src/segy.c
achaikou: I did that for simplicity and flexibility. This code uses just read and seek - so if they are implemented as memread and memseek, it is enough for me.
(And in the other change it actually should be memwrite and memseek, not memread and memseek)

Forcing a flag to me complicates situation a bit as I would have to decide which implementations datasource should have for this flag to apply. Must datasource have memtell? Should it have memclose or is mmapclose good enough?

As in our case answers are obvious, it is doable, but I am not sure if it would be a plus or a minus.
So if you feel strongly about this, I will implement that flag :slightly_smiling_face:

yngve793: You made a nice clean up by using the `segy_datasource` abstraction. The new struct created a well defined interface and resulted in the removal of a significant number of if statements.  I understand that we are breaking this in https://github.com/yngve793/segyio/commit/4abc4a7dbff702f1e75834cd2c282c8f8859a6fb#diff-db14f7535f3fa7555526c5af480c067af64e08149e6854577576856077309be8L2129 and https://github.com/yngve793/segyio/commit/4abc4a7dbff702f1e75834cd2c282c8f8859a6fb#diff-db14f7535f3fa7555526c5af480c067af64e08149e6854577576856077309be8L2284 for the speedup. 

My point is that this should be made visible and not simple. We would like to convey to the reader that of the `segy_datasource` struct that there will be exceptions to this abstraction level. Further more the name of the variable in the struct can be used to describe what we are doing. 

Simple example in https://github.com/equinor/segyio/compare/main...yngve793:segyio:c_datasource_mem_speed

I will not push this any further.

yngve793 (commented):
Added comment.

lib/src/segy.c
achaikou: You've convinced me this is a good idea!

PR #623, opened 2025-05-16
Refs/heads/fix/python binding

yngve793:
Introduce the new binary fields defined SEG-Y_r2.1


python/segyio/field.py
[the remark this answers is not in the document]
yngve793: PO argued to keep C++, Seismic Unix and Matlab in sync without adding additional tests. 
- C++ is updated in commit `Rename binary header fields`
- Seismic Unix words are updated in this commit. 
- Matlab remains. Unclear what needs to be changed. All tests are passing.

python/segyio/su/words.py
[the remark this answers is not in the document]
yngve793: PO argued for keeping seismic unix in sync without adding additional tests.

python/test/segy.py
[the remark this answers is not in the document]
yngve793: Depreciation warning added in fixup.

python/segyio/segyio.cpp
```diff
@@ -1469,23 +1469,86 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
     buffer_guard buffer;
-    int field;
-    int value;
-    if( !PyArg_ParseTuple( args, "w*ii", &buffer, &field, &value ) )
+    if( !PyArg_Parse(buffer_arg, "w*", &buffer) )
         return NULL;
 
-    if( buffer.len() != SEGY_BINARY_HEADER_SIZE &&
-        buffer.len() != SEGY_TRACE_HEADER_SIZE )
-        return BufferError( "buffer too small" );
```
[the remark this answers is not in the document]
yngve793: Yes you are right, it should stay in.

python/segyio/segyio.cpp
```diff
@@ -1435,28 +1435,35 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
+        case SEGY_UNSIGNED_SHORT_2_BYTE:
+            return PyLong_FromLong( fd.value.u16 );
+        case SEGY_UNSIGNED_CHAR_1_BYTE:
+            return PyLong_FromLong( fd.value.u8 );
+
+        case SEGY_IEEE_FLOAT_8_BYTE:
+            return PyFloat_FromDouble( fd.value.f64 );
+
+        default:
+            return KeyError( "No such field %d", field );
```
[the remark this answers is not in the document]
yngve793: Sure, message updated. In setfield fixup.

python/segyio/segyio.cpp
```diff
@@ -1469,23 +1469,86 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
+                    return PyLong_FromLong( fd.value.i32 );
+                case SEGY_SIGNED_SHORT_2_BYTE:
+                    return PyLong_FromLong( (long)fd.value.i16 );
+                case SEGY_SIGNED_CHAR_1_BYTE:
+                    return PyLong_FromLong( (long)fd.value.i8 );
+                case SEGY_IEEE_FLOAT_8_BYTE:
+                    return PyFloat_FromDouble( fd.value.f64 );
+                default:
+                    return KeyError( "Field %d has unknown datatype ", field, fd.datatype );
+            }
```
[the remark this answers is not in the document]
yngve793: Every path in the control flow of the switch statement ends in a return. I do not understand what you want here.

python/segyio/segyio.cpp
```diff
@@ -1435,28 +1435,35 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
+            return PyLong_FromLong( fd.value.i32 );
+        case SEGY_SIGNED_SHORT_2_BYTE:
+            return PyLong_FromLong( fd.value.i16 );
+        case SEGY_SIGNED_CHAR_1_BYTE:
+            return PyLong_FromLong( fd.value.i8 );
+
+        case SEGY_UNSIGNED_INTEGER_8_BYTE:
+            return PyLong_FromLong( fd.value.u64 );
+        case SEGY_UNSIGNED_INTEGER_4_BYTE:
+            return PyLong_FromLong( fd.value.u32 );
```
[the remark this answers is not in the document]
yngve793: Sure, using PyLong_FromUnsignedLong

python/test/segy.py
```diff
@@ -338,10 +338,10 @@ def test_header_dict_methods(openfn, kwargs):
...
         iter(f.header[0])
 
-        assert 35 == len(f.bin.keys())
-        assert 35 == len(list(f.bin.values()))
-        assert 35 == len(list(f.bin.items()))
-        assert 35 == len(f.bin)
+        assert 45 == len(f.bin.keys())
+        assert 45 == len(list(f.bin.values()))
+        assert 45 == len(list(f.bin.items()))
+        assert 45 == len(f.bin)
```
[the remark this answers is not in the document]
yngve793: In Rev 1.02 Unassigned is defined to be 3261. This seams to be a mixup of standards. What do you want to update it to?

python/segyio/segyio.cpp
```diff
@@ -1435,28 +1435,35 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
-            break;
-    }
+    switch ( fd.datatype ) {
 
-    switch( err ) {
-        case SEGY_OK:            return PyLong_FromLong( value );
-        case SEGY_INVALID_FIELD: return KeyError( "No such field %d", field );
-        default:                 return Error( err );
+        case SEGY_SIGNED_INTEGER_8_BYTE:
+            return PyLong_FromLong( fd.value.i64 );
```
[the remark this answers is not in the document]
yngve793: Using long long for 64 bits.

yngve793 (commented):
- Added depreciation warning with tests
- Testing set get datatypes through python
- Addressed comments.

python/segyio/field.py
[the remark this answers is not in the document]
achaikou: To make sure we understand each other (as I am a bit confused by your wording):
I think PO argued not to update Seismic Unix and Matlab to 2.1. As long as we do not break current functionality, we are good.
C++ we plan to update, yes.

python/segyio/segyio.cpp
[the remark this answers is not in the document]
achaikou: `putfield` gets field value from python, goes to c++ and writes value to the file.
Write functions usually do not return anything.

If you exchange whole last SEGY_OK switch case with `return Py_BuildValue("");` (see `putline`, `putdepth`) nothing fails.

You would also get rid of 1) weird casts which are introduced for some unknown reason and 2) reading unsigned values from signed fields, which also baffles me a lot.

python/test/segy.py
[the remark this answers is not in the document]
achaikou: Newest version we support is 2.01, so in our case Unassigned is 3301. 3261 is used already.

So I believe your code `Unassigned1 = 3261` is very much correct, I just suggest to give it a spotlight by putting change into its own commit, so it surely makes it to segyio 2.0 changelog.

python/segyio/segyio.cpp
```diff
@@ -1425,6 +1425,25 @@ PyObject* trbsize( PyObject*, PyObject* args ) {
     return PyLong_FromLong( segy_trace_bsize( sample_count ) );
 }
 
+PyObject* getfieldtype( PyObject*, PyObject *args ) {
```
achaikou: Introducing production function only for test purpose is a very, very questionable approach. I do not think this is a good idea.

And new tests are added to segyio_c and not to segy.py, probably because of this function.
Kind of weird, because functionality itself could be well tested in `segy.py`

python/segyio/segyio.cpp
```diff
@@ -1435,28 +1435,35 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
         return BufferError( "buffer too small" );
 
-    int value = 0;
-    int err = buffer.len() == segy_binheader_size()
-            ? segy_get_field_int( buffer.buf< const char >(), field, &value )
-            : segy_get_field_int(  buffer.buf< const char >(), field, &value )
-            ;
+    segy_field_data fd = segy_get_field( buffer.buf< const char >(), field );
+    if( fd.error != SEGY_OK )
+        return KeyError( "No such field %d", field );
```
achaikou: There could be many reasons for an error here.

Should we properly update `segy_errstr` and return something like `Error( err )` ?

python/segyio/segyformatrange.py
achaikou, line 1: Why is this file added to the segyio library itself??
Would our users need that?

If it a help for tests only, then it should be in the tests directory.

python/segyio/segyformatrange.py
```diff
@@ -0,0 +1,19 @@
+from . import Enum
+
+class SegyFormatRange(Enum):
+    IBM_FLOAT_4_BYTE = None
+    SIGNED_INTEGER_4_BYTE = (-(2**31), 2**31 - 1)
+    SIGNED_SHORT_2_BYTE = (-(2**15), 2**15 - 1)
+    FIXED_POINT_WITH_GAIN_4_BYTE = None
+    IEEE_FLOAT_4_BYTE = None
+    IEEE_FLOAT_8_BYTE = (-(2**50), 2**50)
```
achaikou: Don't know where you got that one from, I think internet says something around 1.80×10**308. So this is kind of a lie, unless you mean test file range, not segy format range.

Anyway, I don't think min-max value matters for floats. If we want to test something specific, it is better to find some 1/(2**x)  number to set as default. But whatever you have is good enough.

python/test/segyio_c.py
```diff
@@ -122,6 +127,76 @@ def get_instance_segyiofd(tmpdir,
...
+    binary_header = f.getbin()
+    for i, (field_name, offset) in enumerate(binfield_keys.items()):
+        if field_name in UNASSIGNED_FIELDS:
+            continue
+        value = _segyio.getfield(binary_header, offset)
+        if field_name in FIXED_VALUE_FIELDS:
+            check_fixed_value_fields(field_name, value)
+        else:
+            range = get_field_range(binary_header, offset, field_name)
+            assert range[0] == value - i, f"Value mismatch for field {field_name}: expected {range[0]} but got {value - i}"
```
achaikou: Back to our eternal discussion as to what should be considered more readable. :joy_cat: 

To me now tests are not easily verifiable. I, a human, simply can't see what is tested.

Two code logics - one that generates test data and one that checks data - have to be aligned here. I have to swim through a lot of code to get to the point of the test.

Current approach might also make it difficult to update field values later as changes would have to fit to the pattern.

To me it would have been a lot easier to just see hardcoded checks instead of this, something very simple like

```
# test/segy.py

@tmpfiles(testdata / 'decrement.sgy')
def test_get_set_binary_header_near_max_values(tmpdir):
    dict = {
        BinField.JobID: 2147483647,
        BinField.EnsembleTraces: 32764,
        BinField.SamplesOriginal: 65526,
        # keep on with other interesting values (e.g. one per equivalence class)
    }
    with segyio.open(tmpdir / 'decrement.sgy', mode='r+', ignore_geometry=True) as f:
        for field, value in dict.items():
            assert f.bin[field] == value
            f.bin[field] -= 1

    with segyio.open(tmpdir / 'decrement.sgy', ignore_geometry=True) as f:
        for field, value in dict.items():
            assert f.bin[field] == value - 1

```

This also drops need for all logic, helper functions, `segyformatrange` and for `getfieldtype`.

python/test/segyio_c.py
```diff
@@ -122,6 +127,76 @@ def get_instance_segyiofd(tmpdir,
...
+            assert value == 1, f"Expected Format to be 1, got {value}"
+        elif field_name == "ExtSamples":
+            assert value == 4, f"Expected ExtSamples to be 4, got {value}"
+        elif field_name == "ExtEnsembleTraces":
+            assert value == 5, f"Expected ExtEnsembleTraces to be 5, got {value}"
+        elif field_name == "ExtendedHeaders":
+            assert value == 0, f"Expected ExtEnsembleTraces to be 0, got {value}"
+
+
+@tmpfiles(testdata / 'increment.sgy')
```
achaikou: You are not writing to a file, so you don't need tmpfiles.
Unless you wish to somehow combine some of the tests to reduce code duplication of boring stuff.

python/test/segyio_c.py
```diff
@@ -122,6 +127,76 @@ def get_instance_segyiofd(tmpdir,
...
+        if isinstance(v, int):
+            datatype_mapping[v] = k
+    return datatype_mapping
+
+def get_datatype_range():
+    datatype_range = {}
+    for k, v in SegyFormatRange.__dict__.items():
+        if isinstance(v, tuple):
+            datatype_range[k] = v
+    return datatype_range
```
achaikou: Those two functions can be local functions inside `get_field_range`.

Doing this really helps to distinguish whether helper functions would be used in many different tests or if they are local to just 1 group of tests.

Would be nice to do the same with two other functions and global variables, but unfortunately I don't see how, unless we create a test class.

python/test/segy.py
```diff
@@ -828,6 +828,23 @@ def test_write_header(small):
         # i.e. don't access buf of treat it as a list
         # assertEqual(list(f.header[2].buf), list(f.header[1].buf))
 
+def test_depricated_fields(small):
+    with segyio.open(small, "r+") as f:
```
achaikou: Why open in `r+` mode? Isn't `r` good enough?

python/segyio/segyio.cpp
```diff
@@ -1435,50 +1454,124 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
-            ? segy_set_field_int( buffer.buf< char >(), field, value )
-            : segy_set_field_int( buffer.buf< char >(), field, value )
-            ;
+    int field = (int)PyLong_AsLong(field_arg);
+
+    segy_field_data fd;
+    int err = segy_init_field_data(field, &fd);
+
+    if( err != SEGY_OK )
+        return KeyError( "No such field %d", field );
```
achaikou: Again - many reasons why this can happen, so we shouldn't disregard `err`.

python/test/segyio_c.py
```diff
@@ -122,6 +127,114 @@ def get_instance_segyiofd(tmpdir,
...
+    binary_header = f.getbin()
+    for i, (field_name, offset) in enumerate(binfield_keys.items()):
+        if field_name in UNASSIGNED_FIELDS:
+            continue
+        value = _segyio.getfield(binary_header, offset)
+        if field_name in FIXED_VALUE_FIELDS:
+            check_fixed_value_fields(field_name, value)
+        else:
+            range = get_field_range(binary_header, offset, field_name)
+            assert range[1] == value + i, f"Value mismatch for field {field_name}: expected {range[0]} but got {value - i}"
```
achaikou: wrong error message: expected range[1], not range[0]

python/segyio/su/words.py
achaikou, line 1: Should we removed all commented  out values as PO doesn't want any new features?

python/segyio/field.py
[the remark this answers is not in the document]
yngve793: Yes, we agree :-)

python/segyio/su/words.py
yngve793: Adding the fields to keep it in sync. My understanding of PO was that we should not put extra effort into updating these bindings, but let the tests pass. Thus, updating the list of fields.

python/segyio/segyio.cpp
yngve793: Sure, I really do not have any idea what you are imagining here. 
Added first draft in new commit. Im an not very happy with the result. This looks like the entrance of a rabbit hole. Suggest to do this in separate PR with a clear understanding of the goal.

python/segyio/segyio.cpp
yngve793: Now returning err as as well.

python/segyio/segyio.cpp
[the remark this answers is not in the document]
yngve793: In segyio the putfield do return the value https://github.com/equinor/segyio/blob/c4f1c012d780083dae3f41dfb0abd41cc504836a/python/segyio/segyio.cpp#L1481

We can change it, but the idea was to not break the interface without a good reason. 

Thanks for pointing out the datatype bug. It was not the intention to read unsigned from signed.

python/test/segy.py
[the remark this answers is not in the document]
yngve793: Is it acceptable to reference this commit directly instead of creating an intermediate step where we have to create a new value that is different from 3301 and 3261?

python/segyio/segyformatrange.py
yngve793: The objective is to test the hole bit range of the field. Floats are a bit tricky as 1 bit is used for sign, 11 for exponent and 52 for precision. https://en.wikipedia.org/wiki/Double-precision_floating-point_format

If we push it beyond 52 bits for precision we will not be able to verify a change of (+/-1). Number 50 was select to have some margin while still pushing boundaries.  Currently there is no test covering exponent bits.

python/segyio/segyio.cpp
yngve793: Lacking access to underlaying datatype is one of the main reason for the extensive refactoring of the repo. Providing the end user with access to this information is in my opinion a completely reasonable thing to do. Note that this will not break any existing code and will not affect efficiency for any existing functions. 

It is possible to redefine everything one more time for testing. Hopefully we can agree that the redefinition of fields and names in different bindings are already causing code duplication and a lot of pain when mantling the code. 
Let us agree on how to solve this before starting to move code. 

We can solve it in a different but:
- I will argue strong for the user to have access to the datatype
- I will argue strongly against increasing the amount of code duplications

python/test/segyio_c.py
yngve793: Moved. In my opinion this is not better.

python/test/segyio_c.py
yngve793: Ok, removed.

python/test/segyio_c.py
yngve793: Given that the code used to generate the tests data has been rejected it is not surprising that you had a hard time here. 

Just because it is differs from what is expected does not mean that is a bad solution. One of the really beautiful properties when testing in this way is how easy it its to generate new reliable test data when adding or changing  features. Hardcoding tests for the entire binary header will generate lots of repeated code and be very resistant to change. It is beyond me why avoiding logic and helper functions is a good thing when testing big and complex objects. 

In my opinion `putfield` and `getfield` can not be tested in a satisfactory way using static test functions.

python/test/segyio_c.py
yngve793: Sure, thanks.

python/test/segy.py
yngve793: Sure, no problem.

python/segyio/segyformatrange.py
yngve793: Moved.

yngve793 (commented):
Thanks, some of the comments where very useful. 
Fixups are added for all comments where I understand the objective and which do not break the test framework. Minor comments:
- Seismic Unix and Matlab we seams to agree in wording but not necessary the implications
- Properly update segy_errstr and return something like Error( err ) need a strategy
- Currently `putfield` return the value. Should we remove this in the Python binding?
- Moved some code. 

Hopefully there is a reasonable way out of this.

python/segyio/su/words.py
achaikou: And I keep being confused because tests pass even with no changes.

And now you even uncommented previously commented out fields?..

Changing deprecated fields is fine, but adding new fields (like `NrTrailerRecords` and others) is unnecessary. 
As we do not plan to support new seismic unix features, why add that??

python/segyio/segyio.cpp
achaikou: OK, I see what you mean.

`Fix return values in Python bindings`  we definitely do not want here. It changes `segy.py` tests, so it means it changes end user behavior. That one we do not want.
`Update segy_errstr in Python binding` would be kinda fine in general, but we do not need it now.

So I agree with you that my suggestion was a bad idea. So scratch all of it.

But see if you instead like keeping the same implementation idea from before this PR, which seems to work:

```
    segy_field_data fd = segy_get_field( buffer.buf< const char >(), field );

    switch( fd.error ) {
        case SEGY_OK: {
            switch( fd.datatype ) {
                    return PyLong_FromLongLong( fd.value.i64 );
                case SEGY_SIGNED_INTEGER_4_BYTE:
                    return PyLong_FromLong( fd.value.i32 );
                // keep on the switch
            }
        }
        case SEGY_INVALID_FIELD:
            return KeyError( "No such field %d", field );
        default:
            return Error( fd.error );
    }
```
So basically handle only invalid field, otherwise return whatever original error was thrown.

python/test/segyio_c.py
```diff
@@ -13,7 +13,24 @@
 import segyio._segyio as _segyio
 from segyio.binfield import keys as binfield_keys
 from segyio.segysampleformat import SegySampleFormat
-from segyio.segyformatrange import SegyFormatRange
+
+class SegyFormatRange(segyio.Enum):
```
achaikou: Still not sure what you wanted to do with floats, but class name is misleading then. If you add a class comment stating that these are ranges used in increment/decrement files which correspond, where possible, to min-max type value, I would be fine with it :cat:

python/segyio/segyio.cpp
achaikou: So the question we should answer first is: would python users ever need the type?

I do not know about any existing usecase for that.
We internally need that type, sure. But end users do not care. They have python `int` and `float` types. They do not need or want anything more. Definitely not for boring stuff like headers. They do not care if field X is originally represented as `uint32` or `int16` as in python they would always get just `int`.

So:
1) Why would users want to have access to internal datatype?
2) What code duplication are you even talking about in this particular case?

python/segyio/segyio.cpp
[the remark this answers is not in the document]
achaikou: Exactly, I suggested changing it in the first comment.
Searching through repo shows that returned value is never used. As nothing needs it, we can safely remove it, because it is not a public interface, but a private one.

---

Maybe we have different opinions about what counts as public user interface?

- C-user interface, that we somewhat allow ourselves to break if truly needed: `segy.h`
- Python end-user interface, that we want to keep backwards compatible: everything that is exported in `segyio/__init__.py`.
- C++ and Matlab interfaces, which we also should try not to break.

C-Python bindings (`segyio.cpp`) is an internal file providing internal interface, which we are allowed to break per our needs.

python/test/segyio_c.py
```diff
@@ -122,6 +144,102 @@ def get_instance_segyiofd(tmpdir,
...
+        if field_name in UNASSIGNED_FIELDS:
+            continue
+        value = _segyio.getfield(binary_header, offset)
+        if field_name in FIXED_VALUE_FIELDS:
+            check_fixed_value_fields(field_name, value)
+        else:
+            range = get_field_range(binary_header, offset, field_name)
+            assert range[0] == value - i, f"Value mismatch for field {field_name}: expected {range[0]} but got {value - i}"
+
+def test_binary_header_datatype_get_dec(tmpdir):
```
achaikou: Both get tests do not longer need tmpdir.
Set ones don't need it too, which is interesting as you change only data in memory, never writing it back to the file. Probably it is even good enough for these tests :smiley:

python/test/segyio_c.py
```diff
@@ -122,6 +144,102 @@ def get_instance_segyiofd(tmpdir,
...
+        value = _segyio.getfield(binary_header, offset)
+        if field_name in FIXED_VALUE_FIELDS:
+            check_fixed_value_fields(field_name, value)
+        else:
+            range = get_field_range(binary_header, offset, field_name)
+            value_set = _segyio.putfield(binary_header, offset, range[1] - i -1)
+            value_get = _segyio.getfield(binary_header, offset)
+            assert value_set == value_get, f"Value mismatch for field {field_name}: set {value_set} but got {value_get}"
+            assert range[1] == value_get+i+1, f"Value mismatch for field {field_name}: expected {range[1]} but got {value_get-i-1}"
+
```
achaikou: Alright, separate question to all 4 new tests:
Why are they implemented in `segyio_c` and not in `segy.py` ?
`segy_py` is testing public functions, `segyio_c` is testing private python bindings.

Also, `f.close()` are missing.

python/test/segyio_c.py
achaikou: Well, then go on with it :+1:  :dove: 

Different solution doesn't mean it is bad, but it has consequences. In this case consequences are related to my well-being, that's why I tried to fight you on it :smile: 
But honestly solution won't really matter in the long run, as we won't interact with this code too often.

---

TLDR; But at least I explained where my difficulties come from :smile: 

Story time. A few years back I implemented a test framework in dlisio that created/verified something in test files automatically. I really liked the beauty of  "how easy it its to generate new reliable test data when adding or changing features" and how it avoided duplication.
Solution got immediately shot down by the project maintainer. With time I accepted some of his reasons for doing that.

> Hardcoding tests for the entire binary header will generate lots of repeated code and be very resistant to change. It is beyond me why avoiding logic and helper functions is a good thing when testing big and complex objects.
- Objects are not complex. This is just a list of values, integers and floats.
- I've made my peace with repeated values in tests long ago. They make test really easy to read, so this trade-off pays off.
- As it has been pointed out to me: values in tests and test data will not change that often. You usually generate it once, then you fix something once or twice, max, and that is all. Yet you will read that test many times, so readability must win.
- Logic: it is easier to compare value 123 with `f.call()` than think "hm. We remove "one" from values that are of different size. Could the logic be flawed? Could it have a bug? Could it down the line result in testing something we don't really want to test?" I am kinda forced to test the tests now. To me it causes aversion to analyze the code, discomfort and inability to make decision on code quality. Bonus: if test ever fails in future or you just want to extend it or are in search of something - you have to repeat the process of understanding the test, again and again.
- My first dislike of helpers here is structure. I really, really, really hate big files as they are hard to have overview of, but previous developers were strongly of a different opinion. To me this file is already hard to read, tests are hard to find, and these helpers make it harder. All you have added would be used just in your 4 tests. It might look better if it all was encapsulated in a class. Currently I just wish for simple tests that are easy to read and distinguish from other tests.
- My second dislike of helpers here is the code that is seemingly irrelevant for the test purpose. For example, I don't need to know about unassigned or fixed fields. I don't really care what is in these fields, but they get their own treatment, as if they are special. From the test purpose point of view all binary header fields are equal. And they are only special here because of the way tests are written.

So in the end it is about what feels good and what makes one uneasy. It is clearly opposite for us now, but lets see, maybe with time our opinions will converge somewhere :)

python/test/segy.py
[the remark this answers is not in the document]
achaikou: It is acceptable of course, I just believe we will forget it by the time we will craft a changelog.

Also maybe I am wrong, but I was under the impression that there won't be any intermediate value? Commit would just change Unassigned1 to 3261 and update values in this test to 36. I don't think anything else will be required?... Otherwise separate commit is of course not worth it.

achaikou (commented):
- Seismic Unix and Matlab: my understanding is "do not touch unless some test complains"
- Agree about Error thing, bad idea, both commits should be removed.
- Yes, I think we can safely remove putfield return value as it is internal and unused!

Of couse we will find a way out of this, it is just a long way out :joy_cat:

python/segyio/segyio.cpp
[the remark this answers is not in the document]
yngve793: Return value is exposed in the Python binding:
>>> f = s.open("../test-data/small.sgy", "r+")
>>> t = f.bin[3213] = 35
>>> print(t)
35

python/segyio/su/words.py
yngve793: Fine, removed.

python/test/segy.py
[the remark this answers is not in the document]
yngve793: Fine, added new commit.

python/segyio/segyio.cpp
yngve793: 1. With this knowledge the user can obtain information such as max and min allowed value and if it is a float. As far as I can remember we already had a case where the value is cased to the target datatype without checking if it fits. 
2. For testing purposes we have to know the datatype to figure out what value to expect. 

I find this discussion a bit baffling. Why do I have to argue for providing the user with completely basic and fundamental information? What is the benefit obtained by withholding this information?

python/test/segyio_c.py
yngve793: Ok, Let us continue this a bit more.

Given your experience with `dlisio` I kind of understand where you are coming from. 

Some history. 
Initially I came from a mindset where tests should be simple, short and ideally just contain the single value we are testing against. At some point I started to work on an existing project where more elaborate test framework was used. It requires more effort to read and manipulate, but had some intriguing additional properties. Examples of these are test for field overlap, offset error around unused block, ensuring that the entire block is read and simple adjustment in case of test file changes. 

Addressing your comments:
- Objects are not complex: It is a mixture of different datatypes with unused blocks scattered in between. For a single field I agree that it is simple but I find the sum complex. 
- Agree, in principle the values should be fixed. In practice they have a tendency to change. 
- Logic: Yes, it is simpler to verify a hardcoded value. On the other hand if the test is very restrictive there should be no need for tests of tests. 
- Big files: Totally agree. The rests should be reorganised into multiple files with descriptive names and maybe even folders. There is no reason to make this harder than it has to be. 
- My second dislike: I am not following you here. I would really like to verify the data in the unassigned blocks, as they are not easily accessible I decided to skip it for now.

python/test/segyio_c.py
yngve793: Ok, fine. Added close.

python/test/segyio_c.py
yngve793: Renamed to SegyFormatTestRange

python/test/segyio_c.py
yngve793: Removed tmpdir

yngve793 (commented):
I am starting to run out of steam here. 

It is hard to keep track at this point. All issues I could find are addressed either by comment of fix.

python/segyio/segyio.cpp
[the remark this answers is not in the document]
achaikou: No, I believe it is not. Search for `putfield` in the repo. I cannot find a single instance of value return from `putfield` not being ignored.

Code you suggest is likely calling
```
        self.putfield(self.buf, key, val)
        self.flush()

        return val
```

python/test/segyio_c.py
achaikou: About moving to `segy.py`
segy.py never directly tests python bindings. It should not do `import segyio._segyio as _segyio`. It tests only public functions that user should have access to.

That's why I first asked you why tests are implemented in segyio_c.

achaikou (commented):
Lets discuss everything that remains in person, so we can agree on most important points and merge this.

python/segyio/segyio.cpp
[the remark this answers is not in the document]
yngve793: Change function signature.

python/segyio/segyio.cpp
yngve793: Now returning err as as well.

python/segyio/segyio.cpp
yngve793: Removed.

python/test/segyio_c.py
yngve793: Tests are removed.

yngve793 (commented):
Without access to field datatype the test framework collapses. 
- Remove tests and related code. 
- Remove return type on `putfield`
- Suggest to skip tests or add them in different PR.

achaikou (approved):
Nice, approved! :cat: 
(you might fix the last commit description (z) and message (heder) if you want, but don't fix if you don't want to)

I suggest we discuss approach to tests together with Alex along with solution to create test data.

achaikou (approved):

PR #624, opened 2025-05-28
Update benchmark test suite

achaikou:
Shall see how this set of tests behaves in future.
- Might reduce number of files, if two files doing the same work just spoils the picture.
- Shall see if file size is good enough. I am not yet sure if relative performance depends on file size or not.
- Together with c-datasource PR `update_by_depth` test was very unstable and `sparse_samples` was certainly slower, but as I slightly updated the criteria (size and how much information is retrieved) , will see how it goes now. Might as well be some weird VM-slowness :woman_shrugging:

### DIFF
--- a/.github/actions/benchmark/action.yaml
+++ b/.github/actions/benchmark/action.yaml
@@ -53,19 +53,19 @@ runs:
       working-directory: python
       run: |
         if [ ! -f file.sgy ]; then
-          time python -m examples.make-file file.sgy 1400 1 1000 1 600
+          time python -m examples.make-file file.sgy 1600 1 1000 1 1200
         else
           echo "file.sgy already exists"
         fi
 
-    - name: Make segy pre-stack file
+    - name: Make segy small file
       shell: bash
       working-directory: python
       run: |
-        if [ ! -f file-ps.sgy ]; then
-          time python -m examples.make-ps-file file-ps.sgy 1400 1 500 1 600 1 3
+        if [ ! -f file-small.sgy ]; then
+          time python -m examples.make-file file-small.sgy 500 1 350 1 350
         else
-          echo "file-ps.sgy already exists"
+          echo "file-small.sgy already exists"
         fi
 
     - name: Print filesize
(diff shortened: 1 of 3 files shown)


.github/actions/benchmark/action.yaml
```diff
@@ -53,33 +53,33 @@ runs:
       working-directory: python
       run: |
         if [ ! -f file.sgy ]; then
-          time python -m examples.make-file file.sgy 1400 1 1000 1 600
+          time python -m examples.make-file file.sgy 1600 1 1000 1 1200
```
yngve793, line 56: Is this generating a larger test file? If so do we want that?

.github/actions/benchmark/action.yaml
```diff
@@ -53,33 +53,33 @@ runs:
...
         ls -lh file.sgy
-        ls -lh file-ps.sgy
+        ls -lh file-small.sgy
 
     - name: Run benchmark tests
       shell: bash
       working-directory: python
       run: |
-        pytest test/benchmarks.py --benchmark-sort=name --benchmark-autosave ${{ inputs.compare }}
+        pytest test/benchmarks.py -rP --benchmark-sort=name --benchmark-autosave ${{ inputs.compare }}
```
yngve793, line 82: Is the purpose of "-rP" to make it more verbose? Will this show in the test outputs in Github actions?

yngve793 (approved):
All tests passes. Code looks fine. 
Only a couple if questions for my understanding.

.github/actions/benchmark/action.yaml
achaikou: Yes, this is a larger file.
Idea is to run performance tests on files relatively similar to the production.
Actual files would be way larger, but we can't create those - github actions does not have a lot of disk memory.
But now as I remove ps file, there is room to increase size of this one.

I am not yet sure if file size matters and how much. For example, overhead of function pointer may not matter for file with 10 000 iterations, but may become visible with 10 mln iterations?

.github/actions/benchmark/action.yaml
achaikou: Yes.
At the moment tests do not print anything, so `-rP` won't have any effect.
However when I am using these tests for checks, it is very convenient have printing options enabled by default to verify workflow does what I think it does.
So this basically spares ma en angry `-rP` fixup "Why do test print nothing in github actions? Oh, that's why... :pouting_cat: "

.github/actions/benchmark/action.yaml
yngve793: Ok

PR #626, opened 2025-06-05
Introduce increment and decrement tests for catb

yngve793:


test-data/decrement.sgy
achaikou, line 1: Should we update Readme.md already now? We can add description of what this file is and leave "recreation" comment empty for now.

applications/test/catb-dec.output
```diff
@@ -0,0 +1,45 @@
...
+extntrpr  	5
+extnart   	2147483619
+exthns    	4
+exthdt    	1125899906842594.000000
+extdto    	1125899906842593.000000
+extnso    	2147483615
+extfold   	2147483614
+intconst  	2147483613
+rev       	218
+revmin    	217
```
achaikou, line 37: These values are `DA` and `D9`, so first bit is unused and these are the only `SEGY_UNSIGNED_CHAR_1_BYTE` values present.

Not sure if we should bother. It is not like revision will ever get this high..

applications/test/catb-dec.output
```diff
@@ -0,0 +1,45 @@
...
+trflag    	32729
+exth      	0
+maxtrh    	32727
+survty    	32726
+timebc    	32725
+ntrst     	-44
+ftroff    	-45
```
achaikou: Those two are unsigned integers, and yet they parsed as signed. :bug:

applications/test/catb-dec.output
```diff
@@ -0,0 +1,45 @@
...
+hstas     	32749
+hstae     	32748
+htatyp    	32747
+hcorr     	32746
+bgrcv     	32745
+rcvm      	32744
+mfeet     	32743
+polyt     	32742
+vpol      	32741
+extntrpr  	5
```
achaikou, line 28: I am totally fine to leave it as is, but just to point out:
might cause confusion as field seems to indicate number of traces in the file, which it does not.

[Makes more sense to have a special value for `ntrst` instead as we might one day read from it.]

test-data/increment.sgy
achaikou, line 1: Absolutely perfect and beautiful files! :heart_eyes_cat:  Number logic is very easy to read :heart: 

Note: not sure how we now deal with number of samples trace header field 115-116 (probably we just ignore it), but it could be updated later if it ever causes troubles on file parse.

applications/test/catb-inc.output
```diff
@@ -0,0 +1,45 @@
...
+rcvm      	-32745
+mfeet     	-32744
+polyt     	-32743
+vpol      	-32742
+extntrpr  	5
+extnart   	-2147483620
+exthns    	4
+exthdt    	-1125899906842594.000000
+extdto    	-1125899906842593.000000
+extnso    	-2147483616
```
achaikou, line 33: Note for future: how interesting that this field overrides nso, but we allow it to be negative.

applications/test/catb-dec.output
yngve793: Well we are NOT testing Segyio here, so I am not sure what signed mean in this context.

applications/test/catb-dec.output
yngve793: First bit is it LSB or MSB?

The purpose of the file is to test the max and min range of the field. As long as the final file is valid the test file should be ok.

applications/test/catb-inc.output
yngve793: Based on the text in the documentation it should probably be an uint32 field.

applications/test/catb-dec.output
yngve793: If `extntrpr ` is set `ntrst ` is free. It is not true the other way around.

test-data/decrement.sgy
yngve793: Let us see how this file will be used first.

applications/test/catb-dec.output
achaikou: Don't understand what you mean (fields talk about different properties), but whatever, doesn't matter that much as long as you yourself are not confused :)

applications/test/catb-dec.output
achaikou: - SEGY_UNSIGNED_CHAR_1_BYTE always has 1 byte, so LSB and MSB does not apply as it applies to the order of bytes in the number only

I assumed that you wanted to test whether signed and unsigned values are also read correctly, that why we go with highest possible values. If highest bit of SEGY_UNSIGNED_CHAR_1_BYTE type is never set, we never check that we do not parse it like SEGY_SIGNED_CHAR, for example.

But as I said, shouldn't matter for revision fields, so ignore if you want.

applications/test/catb-dec.output
achaikou: I think it means that [here](https://github.com/equinor/segyio/blob/705a3f11def08463bde177c452c0e98bd1fc00fa/applications/segyio-catb.c#L161) we should print `llu` and not `lld`

applications/test/catb-dec.output
yngve793: I should probably understand what you want before ignoring it. 

A byte consisting of 8 bits has well defined Least Significant Bit(LSB) and Most Significant Bit(MSB). Are you thinking of endianness?

There is no signed 1-byte field as far as I am aware of. So we will only deal with SEGY_UNSIGNED_CHAR_1_BYTE.

applications/test/catb-dec.output
yngve793: Ok, I will add a fix in separate PR.

applications/test/catb-dec.output
yngve793: From the documentation:
3261–3264 Extended number of data traces per ensemble. If nonzero, this overrides the number of data traces per ensemble in bytes 3213–3214.

achaikou: `ntrst` is bytes 3513–3520 - Number of traces in this file or stream.
"Ensemble" fields can have any value. It is `ntrst` we might, in theory, begin to use later.

But again, I am not asking for any changes, just making sure you are not confused :slightly_smiling_face:

applications/test/catb-dec.output
achaikou: Ohhhhh. Yes, you are right, I though your comment was about endiannes! Good point. 
If you search for lsb and msb in segyio you would see these are used as synonyms to little endian and big endian.
Good that we found out about this confusion. :smile_cat: 

OK, now to the matter.
I thought one of the features of your max range test is to always check how the most significant bit is processed.
`DA` and `D9` do not have the most significant bit set (all other values do).

So with current tests it still can happen that we implemented something badly and value `FF` that should be treated as `SEGY_UNSIGNED_CHAR_1_BYTE` was treated as negative 8-bit signed int instead.

yngve793: 0xD9 = 0b11011001
0xDA = 0b11011010

So most significant bit is set in both cases and least significant is set on D9 case.

achaikou: OK, I am dumb :joy_cat: 
You are right. Forget it :smile:

applications/test/catb-dec.output
yngve793: Ok, let us update the test file when SEGY_BIN_NR_TRACES_IN_STREAM is implemented.

applications/test/catb-dec.output
achaikou: Great! Rebase and fix this and I believe it would be good to merge.

achaikou (approved):
:tada:

PR #628, opened 2025-06-13
Rename entities to reflect broader io support

achaikou:
At the moment we have a lot of entities with name "fd" in it, which likely stands for "file descriptor". As we no longer operate on files, but on an abstract datasource, it is better we rename the classes and variables that have some reference to that.

Alex and I agreed that we want to do some name changes due to having broader io support, but we did not agree on the scope.
I believe proposed changes improve situation, but if reviewer has something against new names, will gladly hear it :cat:

### DIFF
--- a/python/segyio/open.py
+++ b/python/segyio/open.py
@@ -6,7 +6,7 @@ from .utils import c_endianness
 
 def infer_geometry(f, metrics, iline, xline, strict):
     try:
-        cube_metrics = f.xfd.cube_metrics(iline, xline)
+        cube_metrics = f.segyfd.cube_metrics(iline, xline)
         f._sorting   = cube_metrics['sorting']
         iline_count  = cube_metrics['iline_count']
         xline_count  = cube_metrics['xline_count']
@@ -17,7 +17,7 @@ def infer_geometry(f, metrics, iline, xline, strict):
         xlines  = numpy.zeros(xline_count,  dtype=numpy.intc)
         offsets = numpy.zeros(offset_count, dtype=numpy.intc)
 
-        f.xfd.indices(metrics, ilines, xlines, offsets)
+        f.segyfd.indices(metrics, ilines, xlines, offsets)
         f.interpret(ilines, xlines, offsets, f._sorting)
 
     except:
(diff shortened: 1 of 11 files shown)


python/segyio/open.py
```diff
@@ -149,7 +149,7 @@ def open(filename, mode="r", iline = 189,
         raise ValueError(', '.join((problem, solution)))
 
     from . import _segyio
-    fd = _segyio.segyiofd(str(filename), mode, c_endianness(endian))
+    fd = _segyio.segyfd(str(filename), mode, c_endianness(endian))
```
yngve793, line 152: This will be a breaking change in the Python binding. Change internal variable names are perfectly fine. 
Is the change in the python binding intentional?

python/segyio/trace.py
```diff
@@ -102,9 +102,9 @@ class Trace(Sequence):
 
     """
 
-    def __init__(self, filehandle, dtype, tracecount, samples, readonly):
+    def __init__(self, segyfd, dtype, tracecount, samples, readonly):
```
yngve793, line 105: Do we really want to break the binding for a renaming?

>>> import segyio
>>> f = segyio.open("../test-data/small.sgy")
>>> trace_2 = segyio.trace.Trace(segyfd=f, dtype=2,tracecount=4,samples=34,readonly=True)
>>> trace_1 = segyio.trace.Trace(filehandle=f, dtype=2,tracecount=4,samples=34,readonly=True)
Traceback (most recent call last):
  File "<python-input-3>", line 1, in <module>
    trace_1 = segyio.trace.Trace(filehandle=f, dtype=2,tracecount=4,samples=34,readonly=True)
TypeError: Trace.__init__() got an unexpected keyword argument 'filehandle'
>>>

yngve793 (commented):
Renaming to make things more consistent is totality fine. 

Are we sure that these changes should be exposed in the Python binding?

python/segyio/open.py
achaikou: `_segyio` is a python binding and thus considered an internal object.

As per agreement with PO, we care only about our main python interface (exposed via `__init__py`, described in readthedocs, tested in `segy.py` tests)

python/segyio/trace.py
achaikou: Traces are supposed to be created by library and only used by users.
`Trace.__init__` is not declared in the readthedocs, so by our agreement with PO this is considered an internal function. Given that it accepts `_segyfd` internal object, users shouldn't have been using this in the first place.

PR #631, opened 2025-06-27
Expose datasources to python bindings

achaikou:
Python end-user code is planned to be added in the next PR. 
segyio tests will be added in it as I/O methods are not supposed to be called directly from Python, but via other segyio functions. So it must be tested as a whole.

Lack of running examples might make this PR difficult to review though, so commits introducing datasource for users can be added if desirable.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -565,6 +565,11 @@ static int mmapflush( segy_datasource* self ) {
     return SEGY_OK;
 }
 
+static int memflush( segy_datasource* self ) {
+    (void)self; // mark parameter as unused
+    return SEGY_OK;
+}
+
 static int mmapclose( segy_datasource* self ) {
 #ifdef HAVE_MMAP
     int err = mmapflush( self );
@@ -575,6 +580,12 @@ static int mmapclose( segy_datasource* self ) {
     free( mp );
     if ( err != 0 ) return SEGY_MMAP_ERROR;
 #endif // HAVE_MMAP
+    return SEGY_OK;
+}
+
+static int memclose( segy_datasource* self ) {
+    memfile* mp = (memfile*)self->stream;
+    free( mp );
     return SEGY_OK;
 }
 
@@ -672,6 +683,47 @@ segy_file* segy_open( const char* path, const char* mode ) {
 
     ds->minimize_requests_number = true;
     ds->memory_speedup = false;
+
+    return ds;
+}
+
+segy_datasource* segy_memopen( unsigned char* addr, size_t size ) {
+    if( !addr || !size ) return NULL;
+
+    /* as we don't support operations that change file size on update, no memory
+     * resize would be needed
+     */
+    memfile* mp = malloc( sizeof( memfile ) );
+    if( !mp ) {
+        return NULL;
+    }
+    mp->addr = addr;
+    mp->cur = addr;
+    mp->size = size;
+
+    segy_datasource* ds = malloc( sizeof( segy_datasource ) );
+    if( !ds ) {
+        free( mp );
+        return NULL;
+    }
+    ds->stream = mp;
+
+    ds->read = memread;
+    ds->write = memwrite;
+    ds->seek = memseek;
+    ds->tell = memtell;
+    ds->size = memsize;
+    ds->flush = memflush;
+    ds->close = memclose;
+
+    ds->writable = true;
+
+    // on init assume a size of 4-bytes-per-element and big-endian
+    ds->elemsize = 4;
+    ds->lsb = false;
+
+    ds->minimize_requests_number = false;
+    ds->memory_speedup = true;
 
     return ds;
 }
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -735,7 +735,7 @@ int segy_flush( segy_datasource* ds ) {
     if( !ds->writable ) return SEGY_OK;
 
     int flusherr = ds->flush( ds );
-    if( flusherr != 0 ) return SEGY_DS_ERROR;
+    if( flusherr != 0 ) return SEGY_DS_FLUSH_ERROR;
 
     return SEGY_OK;
 }
@@ -759,7 +759,7 @@ int segy_close( segy_datasource* ds ) {
 
     err = ds->close(ds);
     free( ds );
-    if( err != 0 ) return SEGY_DS_ERROR;
+    if( err != 0 ) return SEGY_DS_CLOSE_ERROR;
     return SEGY_OK;
 }
 
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -369,6 +369,12 @@ struct buffer_guard {
 };
 
 struct autods {
+    autods() : ds( nullptr ) {}
+
+    ~autods() {
+        this->close();
+    }
+
     operator segy_datasource*() const;
     operator bool() const;
     void swap( autods& other );
@@ -428,10 +434,8 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
         return -1;
     }
 
-    struct unique : public autods {
-        explicit unique( segy_datasource* p ) { this->ds = p; }
-        ~unique() { this->close(); }
-    } ds( segy_open( filename, mode ) );
+    autods ds;
+    ds.ds = segy_open( filename, mode );
 
     if( !ds && !strstr( "rb" "wb" "ab" "r+b" "w+b" "a+b", mode ) ) {
         ValueError( "invalid mode string '%s', good strings are %s",
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
 }
 
 
+namespace ds {
+
+int py_read( segy_datasource* self, void* buffer, size_t size ) {
+    int err = SEGY_DS_READ_ERROR;
+    PyObject* stream = (PyObject*)self->stream;
+
+    PyObject* result = PyObject_CallMethod( stream, "read", "n", size );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return err;
+    }
+    const char* data = PyBytes_AsString( result );
+    if( data ) {
+        Py_ssize_t result_size = PyBytes_Size( result );
+        if( static_cast<size_t>( result_size ) == size ) {
+            memcpy( buffer, data, size );
+            err = SEGY_OK;
+        }
+    }
+
+    Py_DECREF( result );
+    return err;
+}
+
+int py_write( segy_datasource* self, const void* buffer, size_t size ) {
+    int err = SEGY_DS_WRITE_ERROR;
+    PyObject* stream = (PyObject*)self->stream;
+
+    PyObject* result = PyObject_CallMethod( stream, "write", "y#", buffer, size );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return err;
+    }
+
+    Py_ssize_t result_size = PyLong_AsSsize_t( result );
+    if( static_cast<size_t>( result_size ) == size ) {
+        err = SEGY_OK;
+    }
+    Py_DECREF( result );
+    return err;
+}
+
+int py_seek( segy_datasource* self, long long offset, int whence ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "seek", "Li", offset, whence );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return SEGY_DS_SEEK_ERROR;
+    }
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+int py_tell( segy_datasource* self, long long* pos ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "tell", NULL );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return SEGY_DS_ERROR;
+    }
+    *pos = PyLong_AsLongLong( result );
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+int py_size( segy_datasource* self, long long* out ) {
+    long long original_tell;
+    int err = py_tell( self, &original_tell );
+    if( err != SEGY_OK ) return err;
+
+    err = py_seek( self, 0, SEEK_END );
+    if( err != SEGY_OK ) return err;
+
+    err = py_tell( self, out );
+    if( err != SEGY_OK ) return err;
+
+    err = py_seek( self, original_tell, SEEK_SET );
+    if( err != SEGY_OK ) return err;
+    return SEGY_OK;
+}
+
+int py_flush( segy_datasource* self ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "flush", NULL );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return SEGY_DS_FLUSH_ERROR;
+    }
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+int py_close( segy_datasource* self ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "close", NULL );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return SEGY_DS_CLOSE_ERROR;
+    }
+    Py_DECREF( result );
+    Py_DECREF( stream );
+    return SEGY_OK;
+}
+
+int py_set_writable( segy_datasource* self ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "writable", NULL );
+    if( !result ) {
+        if( PyErr_Occurred() ) {
+            PyErr_Print();
+        }
+        return SEGY_DS_ERROR;
+    }
+    self->writable = PyObject_IsTrue( result );
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+segy_datasource* create_py_stream_datasource(
+    PyObject* py_stream, bool minimize_requests_number
+) {
+    segy_datasource* ds = (segy_datasource*)malloc( sizeof( segy_datasource ) );
+    if( !ds ) return NULL;
+
+    /* current requirements on file-like-object stream: read, write, seek, tell,
+     * flush, close, writable
+     */
+    ds->stream = py_stream;
+
+    ds->read = py_read;
+    ds->write = py_write;
+    ds->seek = py_seek;
+    ds->tell = py_tell;
+    ds->size = py_size;
+    ds->flush = py_flush;
+    ds->close = py_close;
+
+    // writable is set only on init, assuming stream does not change it during
+    // operation
+    const int err = py_set_writable( ds );
+    if( err ) return NULL;
+
+    ds->memory_speedup = false;
+    ds->minimize_requests_number = minimize_requests_number;
+    ds->elemsize = 4;
+    ds->lsb = false;
+
+    /* keep additional reference to assure object does not get deleted before
+     * segy_datasource is closed
+     */
+    Py_INCREF( py_stream );
+    return ds;
+}
+
+} // namespace ds
+
+
 struct buffer_guard {
     /* automate Py_buffer handling.
      *
--- a/python/segyio/segyio.cpp
+++ b/python/segyio/segyio.cpp
@@ -154,44 +154,6 @@ PyObject* Error( int err ) {
     }
 }
 
-struct autods {
-    operator segy_datasource*() const;
-    operator bool() const;
-    void swap( autods& other );
-    int close();
-
-    segy_datasource* ds;
-};
-
-autods::operator segy_datasource*() const {
-    if( this->ds ) return this->ds;
-
-    IOError( "I/O operation on closed datasource" );
-    return NULL;
-}
-
-autods::operator bool() const { return this->ds; }
-
-void autods::swap( autods& other ) { std::swap( this->ds, other.ds ); }
-int autods::close() {
-    int err = SEGY_OK;
-    if( this->ds ) {
-        err = segy_close( this->ds );
-    }
-    this->ds = NULL;
-    return err;
-}
-
-struct segyfd {
-    PyObject_HEAD
-    autods ds;
-    long trace0;
-    int trace_bsize;
-    int tracecount;
-    int samplecount;
-    int format;
-    int elemsize;
-};
 
 struct buffer_guard {
     /* automate Py_buffer handling.
@@ -235,6 +197,45 @@ struct buffer_guard {
     char* buf() const { return this->buf< char >(); }
 
     Py_buffer buffer;
+};
+
+struct autods {
+    operator segy_datasource*() const;
+    operator bool() const;
+    void swap( autods& other );
+    int close();
+
+    segy_datasource* ds;
+};
+
+autods::operator segy_datasource*() const {
+    if( this->ds ) return this->ds;
+
+    IOError( "I/O operation on closed datasource" );
+    return NULL;
+}
+
+autods::operator bool() const { return this->ds; }
+
+void autods::swap( autods& other ) { std::swap( this->ds, other.ds ); }
+int autods::close() {
+    int err = SEGY_OK;
+    if( this->ds ) {
+        err = segy_close( this->ds );
+    }
+    this->ds = NULL;
+    return err;
+}
+
+struct segyfd {
+    PyObject_HEAD
+    autods ds;
+    long trace0;
+    int trace_bsize;
+    int tracecount;
+    int samplecount;
+    int format;
+    int elemsize;
 };
 
 namespace fd {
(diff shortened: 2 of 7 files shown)


python/segyio/segyio.cpp
```diff
@@ -535,7 +543,10 @@ PyObject* close( segyfd* self ) {
     if( !self->ds ) return Py_BuildValue( "" );
 
     errno = 0;
-    self->ds.close();
+    int err = self->ds.close();
```
ajaust: Maybe
```suggestion
     const int err = self->ds.close();
```
since we do not intend to change the `err` variable anywhere. `segyio.cpp` and `segy.c` tend to use const for error return codes where the `err` variable is not reused.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
 }
 
 
+namespace ds {
+
+int py_read( segy_datasource* self, void* buffer, size_t size ) {
```
ajaust, line 160: I am a fan of `const` variables and parameters. Is it possible to make the input `size_t size` constant or must it be mutable?
```suggestion
int py_read( segy_datasource* self, void* buffer, const size_t size ) {
```
My guess is: `PyObject_CallMethod( stream, "read", "n", size )` does call the `read` method of the stream pointer and its signature requires a mutable `size`. Is this guess correct?

I did not dig deep enough to actually understand what `void* stream` is pointing to (the member of the `segy_datasource` struct).

---

More of a style thing that may have an impact on the bindings which I am not aware of: I assume this should be a C++ file, but we use a lot of C-style syntax, e.g., `size_t` over `std::size_t`, `mmap` over `std::mmap` etc. Is this to ensure compatibility with the core code written in C? I would have expected to see more `std::` all over the code.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
 }
 
 
+namespace ds {
```
ajaust, line 158: Do prefer `ds` or `datasource` as name? I assume that `ds` refers to to `datasource`.

I have seen both names and I am wondering which name we should prefer to use for newly added code where we can choose the name.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+        return err;
+    }
+    const char* data = PyBytes_AsString( result );
+    if( data ) {
+        Py_ssize_t result_size = PyBytes_Size( result );
+        if( (size_t)result_size == size ) {
+            memcpy( buffer, data, size );
+            err = SEGY_OK;
+        }
+    }
```
ajaust: Do we have to handle the `else` case here to expose Python's `TypeError` or do we ignore this to trigger our own error handling/reporting?

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+
+    Py_DECREF( result );
```
ajaust, line 180: For my understanding: `result` is a strong reference because we create the object during the `PyObject_CallMethod`. This means we need call `Py_DECREF` such that the reference count is decreased which will lead to the deletion of the object.

Is this somewhat correct?

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+            PyErr_Print();
+        }
+        return err;
+    }
+
+    Py_ssize_t result_size = PyLong_AsSsize_t( result );
+    Py_DECREF( result );
+    if( (size_t)result_size == size ) {
+        err = SEGY_OK;
+    }
```
ajaust: I would consider swapping the call to `Py_DECREF` and the `if` statement for consistency. In the other new methods the call to `Py_DECREF` seems to be the very last call just before the `return` statement of the method.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return err;
+    }
+    const char* data = PyBytes_AsString( result );
+    if( data ) {
+        Py_ssize_t result_size = PyBytes_Size( result );
+        if( (size_t)result_size == size ) {
```
ajaust: Should we use `static_cast` over the C-style case to rule our some implicit. conversion?

```suggestion
        if( static_cast<size_t>(result_size) == size ) {
```
This also applies to the other locations with C-style casts.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+
+    Py_DECREF( result );
+    return err;
+}
+
+int py_write( segy_datasource* self, const void* buffer, size_t size ) {
+    int err = SEGY_DS_WRITE_ERROR;
+    PyObject* stream = (PyObject*)self->stream;
+
+    PyObject* result = PyObject_CallMethod( stream, "write", "y#", buffer, size );
```
ajaust, line 188: ~~What does `"y#"` mean here?~~

Just to confirm that I understand correctly what is going on here:
`PyObject_CallMethod` will call `write` according to the docs:
[PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) *PyObject_CallMethod([PyObject](https://docs.python.org/3/c-api/structures.html#c.PyObject) *obj, const char *name, const char *format, ...)

So `y#` is a format specifier to convert a C string into a bytes array and its size/length [as per docs](https://docs.python.org/3/c-api/arg.html#building-values). I guess it does somewhat the opposite, it takes the buffer and size object and makes it into a C string to call the method write. 

Is this (somewhat correct)?

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+    Py_DECREF( result );
+    if( (size_t)result_size == size ) {
+        err = SEGY_OK;
+    }
+    return err;
+}
+
+int py_seek( segy_datasource* self, long long offset, int whence ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "seek", "Li", offset, whence );
```
ajaust, line 206: `"Li"` means that the parameters are a `long long` and a `int`. Correct? I am gaining confidence in my understanding of the Python interface.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+    ds->write = py_write;
+    ds->seek = py_seek;
+    ds->tell = py_tell;
+    ds->size = py_size;
+    ds->flush = py_flush;
+    ds->close = py_close;
+
+    // writable is set only on init, assuming stream does not change it during
+    // operation
+    int err = py_set_writable( ds );
```
```suggestion
     const int err = py_set_writable( ds );
```

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+    self->writable = PyObject_IsTrue( result );
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+segy_datasource* create_py_stream_datasource(
+    PyObject* py_stream, bool minimize_requests_number
+) {
+    segy_datasource* ds = (segy_datasource*)malloc( sizeof( segy_datasource ) );
+    if( !ds ) return NULL;
```
ajaust, line 292: Here I am (more) confused again about us being in a C++ file (?), but having C-style syntax (`malloc` over `new`, raw pointers instead of smart pointers). I guess this will clarify itself though. 🙂 

Using `NULL` over `nullptr` I think I understand. The pointers are handed to C code and the value of `NULL` is indeed different from `nullptr`.

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+        }
+        return SEGY_DS_SEEK_ERROR;
+    }
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+int py_tell( segy_datasource* self, long long* pos ) {
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "tell", NULL );
```
ajaust, line 219: The format string is `NULL` because the method `tell` does not expect any input parameters. Is this correct?

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+    PyObject* stream = (PyObject*)self->stream;
+    PyObject* result = PyObject_CallMethod( stream, "close", NULL );
+    if( !result ) {
+        if (PyErr_Occurred()) {
+            PyErr_Print();
+        }
+        return SEGY_DS_CLOSE_ERROR;
+    }
+    Py_DECREF( result );
+    Py_DECREF( stream );
```
ajaust, line 270: This is the only method in which the reference counte for `stream` id decreased via `Py_DECREF( stream );`. Why is it not decreased in any of the other methods? Is this due to line 320 (increasing the reference counter such that the stream is not deleted?)?

python/segyio/segyio.cpp
```diff
@@ -155,6 +155,175 @@ PyObject* Error( int err ) {
...
+        }
+        return SEGY_DS_ERROR;
+    }
+    self->writable = PyObject_IsTrue( result );
+    Py_DECREF( result );
+    return SEGY_OK;
+}
+
+segy_datasource* create_py_stream_datasource(
+    PyObject* py_stream, bool minimize_requests_number
```
ajaust, line 289: Use `const`
```suggestion
     PyObject* py_stream, const bool minimize_requests_number
```
or does it save us some copy due to the assignment if we keep `minimize_requests_number` mutable?

python/segyio/segyio.cpp
```diff
@@ -428,10 +434,8 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
         return -1;
     }
 
-    struct unique : public autods {
-        explicit unique( segy_datasource* p ) { this->ds = p; }
-        ~unique() { this->close(); }
-    } ds( segy_open( filename, mode ) );
+    autods ds;
+    ds.ds = segy_open( filename, mode );
```
ajaust: Just for my understanding: What problems did the `unique` struct create that we can avoid by using the C++ mechanism with constructor + destructor?

Jørgen also seemed to be concerned ensuring that the structure is compatibe with C. Is this not of a concern anymore?

python/segyio/segyio.cpp
```diff
@@ -418,39 +418,81 @@ namespace fd {
...
+        "endianness",
+        "minimize_requests_number",
+        NULL
+    };
 
-    if( std::strlen( mode ) == 0 ) {
-        ValueError( "mode string must be non-empty" );
+    if( !PyArg_ParseTupleAndKeywords(
+            args, kwargs, "|ssOip",
+            (char**) keywords,
```
ajaust: Consider following the approach used in `segycreate`. There, a `const_cast< char** >` is used instead of a C-style cast. This would ensure that no other casts are carried out implicitly and it would increase consistency within the source code as both methods would follow the same approach.

python/segyio/segyio.cpp
```diff
@@ -418,39 +418,81 @@ namespace fd {
...
     char* mode = NULL;
-    int endian = 0;
-
-    if( !PyArg_ParseTuple( args, "ssi", &filename, &mode, &endian ) )
-        return -1;
+    PyObject* stream = NULL;
+    int endianness = -1;
+    int minimize_requests_number = -1;
+
+    static const char* keywords[] = {
```
ajaust, line 438: Not sure if this is relevant w.r.t. consistency, but in `segycreate` the list of keywords is named `kwlist`. If this matters, we could consider renaming `keywords` to `kwlist` her as well.

python/segyio/segyio.cpp
```diff
@@ -418,39 +418,81 @@ namespace fd {
...
+        }
 
-    if( !ds && !strstr( "rb" "wb" "ab" "r+b" "w+b" "a+b", mode ) ) {
-        ValueError( "invalid mode string '%s', good strings are %s",
-                mode, "'r' (read-only) and 'r+' (read-write)" );
+        if( std::strlen( mode ) > 3 ) {
+            ValueError( "invalid mode string '%s', good strings are %s",
+                            mode, "'r' (read-only) and 'r+' (read-write)" );
+            return -1;
+        }
```
ajaust, line 497: What safeguard does this check provide over the `!strstr(...)` check in line 474? Could we move the `!strstr(...)` check here and drop it in line `474`?

python/segyio/segyio.cpp
```diff
@@ -418,39 +418,81 @@ namespace fd {
...
         return -1;
     }
 
     if( !ds ) {
-        IOErrno();
         return -1;
     }
 
-    int err = segy_set_endianness( ds, endian );
+    int err = segy_set_endianness( ds, endianness );
```
ajaust: If we decide to add more `const` at the other locations, this is a candidate as well:
```suggestion
    const int err = segy_set_endianness( ds, endianness );
```

python/test/segyio_c.py
```diff
@@ -13,6 +13,10 @@
 import segyio._segyio as _segyio
 
 
+def segyfile(filename, mode, endianness):
+    return _segyio.segyfd(filename=filename, mode=mode, endianness=endianness)
```
ajaust, line 17: Out of curiosity: What motivated this change, i.e., replacing the call to `_segyio.segyfd` with a wrapper function? We still call `_segyio.segyfd` inside the segyio core (`create.py`, `open.py` etc.)

lib/src/segy.c
```diff
@@ -565,6 +565,11 @@ static int mmapflush( segy_datasource* self ) {
     return SEGY_OK;
 }
 
+static int memflush( segy_datasource* self ) {
+    (void)self; // mark parameter as unused
```
ajaust, line 569: Not sure which C standard we use, but in the C23 one could use [`[[maybe_unused]]`](https://www.google.com/search?client=safari&rls=en&q=c23+maybe_unused&ie=UTF-8&oe=UTF-8).

lib/src/segy.c
```diff
@@ -676,6 +687,46 @@ segy_file* segy_open( const char* path, const char* mode ) {
...
+
+    /* as we don't support operations that change file size on update, no memory
+     * resize would be needed
+     */
+    memfile* mp = malloc( sizeof( memfile ) );
+    if( !mp ) {
+        return NULL;
+    }
+    mp->addr = addr;
+    mp->cur = mp->addr;
```
ajaust: Maybe use `addr` twice to make clear that the address and current pointer point to the same address?
```suggestions
	mp->addr = addr;
    mp->cur = addr;
```

lib/src/segy.c
```diff
@@ -676,6 +687,46 @@ segy_file* segy_open( const char* path, const char* mode ) {
     return ds;
 }
 
+segy_datasource* segy_memopen( unsigned char* addr, size_t size ) {
```
ajaust, line 690: Does it make sense to make things more const since we do not plan to change them?
```suggestion
segy_datasource* segy_memopen( unsigned char* const addr, const size_t size ) {
```

lib/src/segy.c
```diff
@@ -676,6 +687,46 @@ segy_file* segy_open( const char* path, const char* mode ) {
...
+    ds->tell = memtell;
+    ds->size = memsize;
+    ds->flush = memflush;
+    ds->close = memclose;
+
+    // on init assume a size of 4-bytes-per-element and big-endian
+    ds->elemsize = 4;
+    ds->lsb = false;
+
+    ds->writable = true;
```
ajaust: Move the assignment of `writable` above the assignment of `elemsize` to ensure same order of initialisation as the members of the `segy_datasource` struct are defined. This a consistency and readability thing.

python/segyio/segyio.cpp
```diff
@@ -381,6 +381,9 @@ struct autods {
     int close();
 
     segy_datasource* ds;
+    // used only for memory datasource. Py_buffer's lifetime must be the same as
+    // of segyfd
+    Py_buffer memory_ds_buffer;
```
ajaust, line 387: `ds` is a pointer to its type, but `memory_ds_buffer` is not. Why do we prefer the pointer for `segy_datasource`, but not for `Py_buffer`?

python/segyio/segyio.cpp
```diff
@@ -392,13 +395,21 @@ autods::operator segy_datasource*() const {
...
+
 int autods::close() {
     int err = SEGY_OK;
     if( this->ds ) {
         err = segy_close( this->ds );
     }
     this->ds = NULL;
+    if ( this->memory_ds_buffer.buf ) {
+        PyBuffer_Release( &this->memory_ds_buffer );
+        this->memory_ds_buffer = Py_buffer();
```
ajaust, line 412: Why do we need to assign a new buffer after releasing the buffer?

python/segyio/segyio.cpp
```diff
@@ -458,6 +472,14 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
         ds.ds = ds::create_py_stream_datasource(
             stream, minimize_requests_number
         );
+    } else if( memory_buffer ) {
+        auto flags = PyBUF_CONTIG | PyBUF_WRITABLE;
+        if( PyObject_GetBuffer( memory_buffer, &ds.memory_ds_buffer, flags ) != 0 )
```
ajaust: We tend to follow the approach to 
1. Carry out some action
2. Check for error
among this file and also the C code.

This line currently does both in the same line. Maybe we should split into something like this:
```suggestion
        const int err = PyObject_GetBuffer( memory_buffer, &ds.memory_ds_buffer, flags );
        if( err != 0 )
```

python/segyio/segyio.cpp
achaikou: ds is a shortcut, datasource is a full name.
So I tend to use `ds` in variable names and, like here, in namespace, to keep up with current code style which prefers shortcuts. I prefer `datasource` in type names where it doesn't create some ill-fitting name.

python/segyio/segyio.cpp
achaikou: No, it is not required by python call.
Signature was created by us. This function implements [function pointer](https://github.com/equinor/segyio/blob/5287707be4c50103eae58b74291bd6686bf3698d/lib/include/segyio/segy.h#L57).

After discussion agreed to leave it const-less for consistency. Current segy.c code doesn't add const modifiers to function parameters.

---

There are no "mmap" calls in segyio.cpp, but `size_t` is coming from function pointer, which is segy.c.

python/segyio/segyio.cpp
achaikou: If err is not set to `SEGY_OK,` default read error would be returned, so we trigger our own error.

As I have no idea what should happen for `data` to become False, I think we better deal with it once we get the actual case.

python/segyio/segyio.cpp
achaikou: Yes, that is also my understanding.
I understood it that if some function does not a return new reference that must be dealt with, it would have been mentioned in the docs.

Decreasing reference will not directly lead to deletion of object (there might be other references), but keeping the reference will definitely prevent object from being garbage collected.

python/segyio/segyio.cpp
achaikou: Yes.

It would take buffer and size, which are c concepts, and create a Python object - `pybuffer` argument for python `def write(self, pybuffer)` function.

python/segyio/segyio.cpp
achaikou: Correct!

> The format can be NULL, indicating that no arguments are provided.

There are some versions of the method that use "no args", but I believe I cannot use them just by stating the name of the python method.

python/segyio/segyio.cpp
achaikou: Exactly.

We increased the stream reference once on init, now we decrease it once in `close` as we don't bother keeping stream alive anymore.

If we decreased it before `close`, stream object might have got deleted without our knowledge.

python/segyio/segyio.cpp
achaikou: Agreed not to add const to function parameters

python/segyio/segyio.cpp
achaikou: Those functions were copied from C code (see segy_open and segy_mmap) and adjusted.  I didn't think of rewriting them to c++ and consider this whole `ds` namespace to be more C-Python than C++-Python.

I see your point of using c++ features here, but I slightly lean into leaving code similar between all datasource creating functions.

---
Haven't actually registered the NULL :smile:
It is used over the whole file, so I probably was unaware and copied it from somewhere :smile:

python/segyio/segyio.cpp
achaikou: I wasn't able to decipher what he was concerned about...
My best guess was constructor/destructor behavior, so I tried explaining in the commit message why I think new code is ok.

----
Problem with unique were the next commits where I introduce other datasources. Maybe it is my bad c++ showing, but I was struggling for a long time with writing it nicely... 

For example, `ds` is one object with one _defined_ non-empty constructor, but there are different `file + stream + memory` ways of creating it, which resulted in clumsy code.

Especially the memory bit, where I was struggling with the design of keeping original python buffer alive (so without closing or copying it). In the working "unique" solution there were move constructors/assignments in all 3 nearby classes. Don't remember though if getting rid of `unique` helped getting rid of them as well or it was unrelated.

And I couldn't understand the real purpose of `unique`. At all. It all began to look nice and somewhat understandable to me only after I deleted the class.

python/segyio/segyio.cpp
achaikou: :open_mouth: 
I haven't even noticed we have the same thing in `segycreate` :smile: 

To me `segyopen`/`segycreate`/`suopen` are the ones where internal consistency would matter somewhat. `init` is a bit aside as it always happens in the first step, followed by any of the previous 3.

Besides, I like `keywords` more than `kwlist` :smile:

python/segyio/segyio.cpp
achaikou: It says

> Remove the const_cast when python2 support is dropped

So I am not sure what was going on there. I added that cast because some analyzer was complaining.

Changed to `const_cast` as per agreement.

python/segyio/segyio.cpp
achaikou: I didn't intend to change file handling logic in this commit, so didn't particularly think about it.

I suspect this has something to do with first check happening before file open and another one after and a [few checks](https://github.com/equinor/segyio/blob/5287707be4c50103eae58b74291bd6686bf3698d/lib/src/segy.c#L618) in the function itself. There might be some kind of reasoning behind that.

If we figure it out, then probably not in this PR.

python/test/segyio_c.py
achaikou: ~Laziness~ Switching from positional arguments to keywords :smile: 

I decided it is easier to have a wrapper here, than to add `filename = ..., mode = ..., endianness = ...` on every call in test

lib/src/segy.c
achaikou: Good to know, thanks!

We definitely don't use C23 here. 

> 
> To build segyio you need:
> 
>     A C99 compatible C compiler (tested mostly on gcc and clang)

Didn't see anything worth upgrading yet, but as we do not require any standard in Cmake, no idea what we actually build with in CI.

lib/src/segy.c
achaikou: How you love your consts :smile: :heart: 

Left `size_t` as is due to current style.
`unsigned char* const` I understand also would refer to non-modifiable pointer. Current code also doesn't practice that.
I believe current code supports const only in "do not modify the data itself" meaning.

python/segyio/segyio.cpp
achaikou: Memory management was one of my biggest problems here :smile: 

`segy_datasource` objects are created with malloc/calloc in C `segy_open`, `segy_mmap`, etc functions. They then are passed absolutely everywhere in `segy.c` and `segy.cpp` , obviously through pointers.

`Py_buffer` object is not passed anywhere. Object is created once, lies forgotten, and then is destroyed at the end. Its only purpose is to live - so that it's member, `.buf` char pointer, can be safely used.

So we could either create and keep it directly, like here, or on heap with `std::unique_ptr` (or something). But I saw no reason to introduce additional layer of pointer/heap here as object is unused anyway.

Would prefer to keep it simple, unless you see benefits of some other solution. :cat:

python/segyio/segyio.cpp
achaikou: We release the old buffer from memory - we close the instance.
Now we kinda want to "nullify" the field by creating a default empty instance in its place, so the old one is definitely unreachable. Similar to assigning `this->ds = NULL;` 3 lines above.

Don't remember if that affected any of my tests though.

python/segyio/segyio.cpp
achaikou: I see that I copied that code from `buffer_guard` constructor, so we definitely have a mix of approaches there, but your suggestion makes more sense :+1:

achaikou (commented):
Thanks for the review! :cat:

python/segyio/segyio.cpp
ajaust: Just to add what I interpreted into Jørgens commit, but not guarantee that it is correct:

1. Compatibility (with C) meant that there are not constructors or destructors in the struct `autofd` because C does not have this concept.
2. The `unique` struct was a way to wrap a C object that cleans up after itself up by wrapping the object into a C++ structure with a constructor and destructor. By extending `autofd` he could use it as the "normal" `autofd` within the `init` method. When `fd` goes out of scope, the object is destroyed and thus automatically closes the file descriptor.

python/segyio/segyio.cpp
ajaust: I don't know either, but I would have interpreted it as: "In Python3 we can keep the `const` and don't have to cast it away". 

The [Python docs](https://docs.python.org/3/c-api/arg.html#c.PyArg_ParseTupleAndKeywords) mention that it supports quite some constness in the keywords parameter. If the analysers complain, we can keep this and revisit it sometimes later.

python/segyio/segyio.cpp
ajaust: Sounds good. I also like simple solutions. 🙂

PR #632, opened 2025-07-03
Expose datasource to python

achaikou:
Datasource finally reaching the users. :checkered_flag: 
Please double check that I am providing them with enough correct and useful information :innocent: and that this solution is good enough and works as I think it does :sweat_smile: .

Only thing left is benchmark workflow, but I want to add it after this PR is merged.


python/segyio/segy.py
```diff
@@ -23,13 +23,16 @@ class SegyFile(object):
...
-                                           xline=193,
-                                           endian='big',
-                                           ):
-
-        self._filename = filename
-        self._mode = mode
+    def __init__(
+        self,
+        fd,
+        datasource_descriptor,
```
ajaust, line 29: I guess it becomes clearer later on, but my first thought here is/was "Why do I need two descriptors here now (`fd` + the datasource descriptor) when one was enough before?"

python/segyio/create.py
```diff
@@ -181,6 +184,38 @@ def create(filename, spec):
...
+
+    Parameters
+    ----------
+
+    stream : file-like object
+        Data destination. It is up to the user to assure stream is opened in w+b mode.
+    spec : segyio.spec
+        Structure of the segy file
+    minimize_requests_number : bool
+        Configuration for certain algorithms. True to minimize number of
```
ajaust: "certain" raises the questions "which ones" for me. Maybe rephrase it slightly to something like "Configration of internal algorithms" or similar?

python/segyio/create.py
```diff
@@ -181,6 +184,38 @@ def create(filename, spec):
...
+        requests to the stream at the cost of reading/writing unnecessary data.
+        False to minimize memory usage at the cost of many requests.
```
ajaust: I am slightly confused that the parameter set to `True` and `False` is not configuring exactly opposite for the memory/data part.

- `True` reads and writes unnecessary data if set to `True`, i.e., I guess it means more data is written than strictly required to update some fields. Or is it indeed unnecessary data in the sense of data that does not make sense/is bloat? Does it mean more data on the storage is used or more of the I/O or bandwidth is used for operations that one does not need?
- `False` reduces the memory usage. However, memory usage is something else to me than writing unnecessary data.

Am I understanding it wrong? To me, memory usage refers to how much memory/RAM is necessary.

python/segyio/utils.py
```diff
@@ -65,3 +65,52 @@ def make_segyfile_descriptor(self, endian):
             endianness=c_endianness(endian)
         )
         return fd
+
+
+class StreamDatasourceDescriptor():
+    def __init__(self, stream, minimize_requests_number):
+        if stream is None:
+            raise ValueError("stream must not be None")
```
ajaust: Is this error message in line with typical segyio error messages or should we be more descriptive? At the moment the error message is quite close to the default exception message if we would just skip the check and wait for the exception once we try to use the object. Would something like "No stream is passed into the data source" be suitable?

Should we additionally check for a certain type/methods to be part of the `stream` object? We could check whether the stream has all the methods that we expect, but I guess what methods we need also depends on what the user is going to do (only reading vs read and write). The initialiser does not have default arguments so `stream` should ever be `None` if the user is accidentally passing something that is `None` or if they explicitly pass `None` into the method.

python/segyio/open.py
```diff
@@ -154,6 +158,76 @@ def open(filename, mode="r", iline = 189,
...
+    Parameters
+    ----------
+
+    stream : file-like object
+        Source of the data. It is up to the user to assure stream is opened in rb
+        mode for reading and r+b mode for updating.
+    minimize_requests_number : bool
+        Configuration for certain algorithms. True to minimize the number of
+        requests to the stream at the cost of reading/writing unnecessary data.
+        False to minimize memory usage at the cost of many requests.
```
ajaust: If we update the docstring in `create_with` this should be updated accordingly.

python/segyio/open.py
```diff
@@ -154,6 +158,76 @@ def open(filename, mode="r", iline = 189,
...
+
+
+def open_in_memory(memory_buffer,
+                   iline=189,
+                   xline=193,
+                   strict=True,
+                   ignore_geometry=False,
+                   endian='big'):
+    """
+    Opens a segy file from memory.
```
ajaust, line 208: The method name says `open_in_memory`, but the docstring says `from memory`. I think it should be `in` or `into` in the docstring.

python/segyio/open.py
```diff
@@ -154,6 +158,76 @@ def open(filename, mode="r", iline = 189,
...
+
+    Function behaves the same as `segyio.open`, but expects file to fit fully
+    into memory. Performance is expected to be significantly faster than using
+    `segyio.open` or `segyio.open_with`.
```
ajaust: I might like to have an even more explicit statement that the file will be fully loaded into memory just so users do not expect any lazy loading or similar.

It is mentioned that it is expected that the file fits fully into memory, but it is not stated what happens if it does not fit. Is it worth mentioning what is going on?

Maybe we could change the docstring to something like this:
> Function behaves the same as `segyio.open`, but loads the file fully into memory. If the file does not fit into memory XXX happens. Performance is expected to be significantly faster than using`segyio.open` or `segyio.open_with`.

python/segyio/open.py
```diff
@@ -154,6 +158,76 @@ def open(filename, mode="r", iline = 189,
...
+
+    Parameters
+    ----------
+
+    buffer : contiguous memory object
+        Source of the data: complete SEGY file loaded into memory. Update
```
ajaust, line 218: We use `fully` in the docstring a few lines above, but say `complete` here. Could we make the description more consistent to use the same style in the docstring summary as in the parameter description? I think this would make it easier for users to understand what is going on.

python/segyio/utils.py
```diff
@@ -65,3 +65,52 @@ def make_segyfile_descriptor(self, endian):
...
+            endianness=c_endianness(endian),
+            minimize_requests_number=self.minimize_requests_number,
+        )
+        return fd
+
+
+class MemoryBufferDatasourceDescriptor():
+    def __init__(self, buffer):
+        if buffer is None:
+            raise ValueError("buffer must not be None")
```
ajaust: If we consider updating the error message in `StreamDatasourceDescriptor` this is also candidate to be rephrased.

.github/workflows/build.yaml
```diff
@@ -134,3 +134,31 @@ jobs:
...
+        shell: bash
+        run: |
+          cmake -S . -B build -DBUILD_PYTHON=ON -DBUILD_BIN=OFF
+          cmake --build build
+
+      - name: Test
+        shell: bash
+        run: |
+          cd python
+          pytest -rP --datasource ${{ matrix.stream }} test/stream.py
```
ajaust, line 164: Not an issue for this PR, but something I realised when digging into the CMake stuff.

What is our preferred way of triggering tests? We have up to two options depending on how segyio is build: 

1. CMake
2. pytest

If we build segyio purely from CMake with `DBUILD_PYTHON=ON` we tend to run the tests via `ctest`. We do not do this in the `large` tests. The tests run by `ctest` include (some of?) the Python tests.

If we build the segyio Python library/bindings then we use `pytest`. This make sense since we don't necessarily need to build the segyio core library via the CMake build step.

My impression is that the `DBUILD_PYTHON=ON` is/was introduced especially for the developers who build the library on their machines via CMake and then run the tests via `ctest` often. It could make sense to have some stream tests in there.

---

More relevant to this PR:
Why is this a separate job within the action instead of integrating the test into the CTest suite? 

Should this test be added to the CTest suite? Check the the bottom of the `CMakeLists.txt` in `python/` to find the defined tests.

python/setup.cfg
```diff
@@ -6,4 +6,4 @@ test=pytest
 
 [tool:pytest]
 python_files = test/*.py
-addopts = --ignore=test/benchmarks.py --ignore=test/large.py
+addopts = --ignore=test/benchmarks.py --ignore=test/large.py --ignore=test/stream.py
```
ajaust, line 9: Why is the test excluded from the "normal" tests? I see that benchmarks and the large tests are also inside their own GitHub Action, but this is not the case for the stream.

test-data/README.md
```diff
@@ -25,6 +25,7 @@ Due to changes in the scripts after the files were originally created, reproduci
 | small.sgy                      | Inline-sorted file with 50 samples, 5 inlines-xlines test file.             | make-file.py small.sgy 50 1 6 20 25                   | Basic file used in testing                 |
 | text-embed-null.sgy            | small.sgy but with one byte in text header exchanged with 00                |                                                       |                                            |
 | text.sgy                       | File with unique textual file header.                                       |                                                       | no traces                                  |
+| tiny.sgy                       | Inline-sorted file with 4 samples, 3 inlines, 2 xlines.                     | make-file.py small.sgy 4 1 4 20 22                    |                                            |
```
ajaust, line 28: It is mentioned in the commit message, that `tiny.sgy` is needed for the cloud testing. However, it is not 100% clear to me why we cannot use `small.sgy` for example and what. What problems/limitations do we run into if we use an existing file like `small.sgy`? Is there any indication for how much faster/slower it is to use `small.sgy` vs `tiny.sgy?

Is it possible to motivate the need for `tiny.sgy` a bit more in the commit message?

python/test/conftest.py
```diff
@@ -0,0 +1,161 @@
+import pytest
+
+import os
+import io
+
+import shutil
```
ajaust, line 6: Do we care about alphabetical ordering of the imports in segyio?

python/test/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def get_filename_and_filepath(path):
```
ajaust, line 9: Out of curiosity: Is there any obvious choice here to decide between `pathlib` and `os.path` for our use case? I was thinking whether it could look neater with `pathlib`, but I have no idea.

python/test/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+        filename = path
+        filepath = os.path.abspath("../test-data/"+path)
+        return filename, filepath
+
+    # assuming file lies in current working directory
+    return path, os.getcwd() + "/" + path
+
+
+@pytest.fixture()
+def make_nothing_fixture():
```
ajaust, line 25: Do we prefer the `_fixture` suffix as part of the method's name? We have already marked it as fixture via the decorator.

python/test/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+    def _make_nothing(*args, **kwargs):
+        return None
+    yield _make_nothing
+
+
+@pytest.fixture()
+def make_python_file_stream_fixture(tmp_path_factory):
+    """
+    Creates steam from a standard python file.
```
ajaust: steam -> stream

python/test/conftest.py
```diff
@@ -0,0 +1,161 @@
...
+
+        buffer = MemoryIO(open(filepath, mode).read())
+        buffers.update({filename: buffer})
+        return buffer
+    yield _make_stream
+
+
+@pytest.fixture()
+def make_memory_buffer_fixture():
+    """ Creates memory buffer from local file."""
```
ajaust:
```suggestion
     """Creates memory buffer from local file."""
```

.github/workflows/build.yaml
```diff
@@ -134,3 +134,31 @@ jobs:
...
         with:
           command: addpath('build/mex'), try, run('build/mex/test/segyspec.m'), exit(0), catch ME, disp(getReport(ME, 'extended')), exit(-1), end;
+
+  python:
+    runs-on: ubuntu-latest
+    name: Build and test python on example stream
+    strategy:
+      fail-fast: false
+      matrix:
+        stream: [pyfile, pymemory]
```
ajaust, line 144: We also define the option `nativememory`. Do we need any test for this?

python/test/memory_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_read():
+    with open(small, "rb") as f:
+        data = bytearray(f.read())
+
+    with segyio.open_in_memory(data) as f:
+        cube = segyio.tools.cube(f)
+
+    with segyio.open_in_memory(data) as f:
+        assert f.tracecount == 25
```
ajaust: What is the benefit over this:
```suggestion
    with segyio.open_in_memory(data) as f:
        cube = segyio.tools.cube(f)

        assert f.tracecount == 25
```
Do we explicitly want the file to be closed before opening it for the assertions?

python/test/memory_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+
+@pytest.fixture
+def memory_source():
+    def setup_memory_source(data):
+        def open_source(mode="r"):
+            return segyio.open_in_memory(data)
+        return open_source
+    return setup_memory_source
```
ajaust: I am surprised to see so many layers of methods, but I am sure it makes sense. 😁

python/test/memory_buffer.py
```diff
@@ -0,0 +1,79 @@
...
+
+def test_reopen_closed():
+    with open(small, "rb") as f:
+        data = bytearray(f.read())
+
+    with segyio.open_in_memory(data) as f:
+        pass
+
+    with segyio.open_in_memory(data) as f:
+        pass
```
ajaust: ~~Are we doing this twice on purpose or is this a copy and paste error?~~

~~If it is done on purpose, then I do not understand yet why.~~

So we open the data in memory once, don't do anything such that it is closed and then open it again to see if it works. Is this correct? I was first confused that I did not see any `close()` anywhere.

python/test/segy.py
```diff
@@ -37,10 +37,25 @@
                        'endian': 'lsb' }),
 ]
 
+def load_contents(filepath):
+    with open(str(filepath), "rb") as f:
+        data = bytearray(f.read())
+    return data
```
ajaust, line 43: I was first confused why the definition of SEGY files and SU files are split by some method definition, but I see now that the method is used within the definiion of small SEGY files.

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
 
     with pytest.raises(KeyError):
         _segyio.fread_trace0(25, 1, 1, 1, indices, "depth")
+
+
+@pytest.mark.parametrize("datasource", ["stream", "memory"])
+def test_datasource_little_endian(datasource):
+    if datasource == "stream":
+        stream = open(testdata / "small-lsb.sgy", "rb")
```
ajaust, line 601: Do we need to close this stream or is it closed automatically when stream goes out of scope? I am just reacting here since we do not open the stream using the `with open` syntax.

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
...
+
+    fd.close()
+
+
+def test_memory_buffer_memory_management():
+    with open(testdata / "small.sgy", "rb") as f:
+        data = bytearray(f.read())
+    fd = _segyio.segyfd(memory_buffer=data, endianness=0).segyopen()
+    # assure test code owns no "data" reference, so it can be garbage collected
+    del data
```
ajaust, line 627: Is it good enough to not have a memory reference to `data` or do we need to be sure that that is garbage collected before continuing?

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
...
+
+    # trying our best to accidently overwrite memory actually used by "data"
+    _ = [numpy.zeros(250, dtype=numpy.single) for _ in range(100)]
```
ajaust: How likely is it to actually overwrite the one memory location that we actually read in the end? Should we maybe try to hash the full buffer to see if any field in the buffer has changed?

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
...
+    fd.close()
+
+
+@pytest.mark.xfail(
+    reason=(
+        "Test assures users get an exception, not a segfault, "
+        "so xfail is intentional."
+    ),
+    strict=True
+)
```
ajaust, line 647: Nice to see tests that check for error handling.

Would it have any benefit to explicitly expect a [certain exception](https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions)? This would ensure that the exception that you expect is thrown instead of some other exception that may be cause by something else or an exception that is less informative than what the user should see.

python/test/stream.py
```diff
@@ -0,0 +1,232 @@
+import segyio
+import numpy as np
+import pytest
```
ajaust, line 3: Somewhat different import order than in `conftest.py`, but I am not sure if we care.

python/test/stream.py
```diff
@@ -0,0 +1,232 @@
...
+
+testfile = "tiny.sgy"
```
ajaust, line 5: In `python/test/memory_buffer.py` the small file variable is called `small`, but here it is `testfile`. Not sure if we should care though.

python/test/stream.py
```diff
@@ -0,0 +1,232 @@
...
+    for iline_index, iline in enumerate(f.ilines):
+        iline_data = f.iline[iline]
+        for xline_index, xline in enumerate(f.xlines):
+            trace = iline_data[xline_index]
+            tracenr = iline_index * len(f.xlines) + xline_index
+            if tracenr % trace_stride == 0:
+                chosen_samples = next(chosen_samples_generator)
+
+            for sample_index in range(len(f.samples)):
+                expected = iline + xline / 100.0 + sample_index * step
```
ajaust, line 82: Is the `100.0` a magic number that strems from how the file `tiny.sgy` is set up or is this a constant that is valid for all SEGY files?

python/test/stream.py
```diff
@@ -0,0 +1,232 @@
...
+        with open_with_stream(make_stream, fresh) as f:
+            assert f.bin == src.bin
+            assert np.array_equal(f.header, src.header)
+            assert np.array_equal(f.trace, src.trace)
+
+
+def run_test_update(open_datasource):
+    """
+    Aims to call all main internal paths that update data and ensure correct results.
+    Works only on test files created through "make_file" example.
```
ajaust, line 149: Why is it mandatory for files to be created by `make_file`?

python/test/stream.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+@pytest.mark.xfail(
+    reason=(
+        "Test assures users must get an exception, not a segfault, so xfail is intentional. "
+        "Didn't always appear on first execution, so test is repeated a few times. "
+        "Test was sponsored by one missing Py_INCREF"
+    ),
+    strict=True
+)
```
ajaust, line 227: If we consider it useful in `python/test/segyio_c.py` to check for a specific exception rather than just an any exception, we should also consider it here.

python/tutorials/using_own_stream.ipynb
```diff
@@ -0,0 +1,393 @@
...
+    "   pytest.main([\"--custom_stream\", \"--durations=5\", \"../test/stream.py\"], plugins=[plugin])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7520b45c",
+   "metadata": {},
+   "source": [
+    "### Running tests on local file\n",
+    "Note that running `segyio.open_with` on local file obtained through python's own `open(filename)` is slower than running `segyio.open(filename)` ."
```
ajaust: Here, I would personally prefer capitalisation of "python's" -> "Python's" as it is used a name.

```suggestion
    "Note that running `segyio.open_with` on a local file obtained through Python's own `open(filename)` is slower than running `segyio.open(filename)`."
```

python/tutorials/using_own_stream.ipynb
```diff
@@ -0,0 +1,393 @@
...
+   "execution_count": null,
+   "id": "d89a8464",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Define your plugin class which implements 'make_stream' fixture.\n",
+    "# Testing framework will call 'make_stream(path, mode)'.\n",
+    "#\n",
+    "# Note that some tests will make a call several times to the same path.\n",
+    "# It is up to the user to assure same I/O is returned.\n",
```
ajaust: What das "same I/O" mean here? I am not sure what "I/O" and "same" mean. Maybe we should explain here:

1. When are two "I/O"s are the same.
2. What do we understand an "I/O" object is or rather what properties it must fulfill? Is it a file-like object or something else? I think up to here we use "I/O" in a more generic context where one does not need to know this.

python/tutorials/using_own_stream.ipynb
```diff
@@ -0,0 +1,393 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "ef084bf1",
+   "metadata": {},
+   "source": [
+    "We can see that `test_create` and `test_update` failed as functionality is not supported by this I/O. That informs us that this I/O is compatible only with segyio read functionality.\n",
+    "\n",
+    "We also see that reading is very slow compared to local file, which is to be expected."
```
ajaust, line 294: From a perspective of someone that is new to segyio and cloud storage: Do we really see that reading on the local disk is slow? There are no numbers for local file access reported. Could we maybe state a number of what we would normally expect for file accesses?

python/tutorials/using_own_stream.ipynb
ajaust, line 1: I like this tutorial. I think it will be very helpful for users testing out this feature. I did not fully run the notebook on my machine yet. Let me know if I should test it out myself.

python/segyio/segy.py
achaikou: Best name I could come up with after lots of struggles and renaming... :unamused: 

`fd` is SegyFile descriptor. It is common for all datasources because they share the same SegyFile
`datasource_descriptor` is descriptor of all the stuff that is difference in each datasource...

Originally I extended SegyFile and had stuff like FileSegyFile, StreamSegyFile, but it didn't look good with regards to users being used to only one SegyFile. So I had to put all the different information to some composite class instead. Datasource_descriptor was the best name I found...

python/segyio/create.py
achaikou: I wanted to give some lightweight pointers to the users on what parameter might affect, without in-deep explanation, given that default value is supplied.

I have a trace with 1000 4B samples. I want to add 10 to every 50th sample.

`minimize_requests_number = True`
- We read the whole trace into memory in one request.
- We update every 50th sample
- We write whole trace to the file in one requests.

Total: 2 requests, 1000 x 4B of memory used. Was originally used for File path.

`minimize_requests_number = False`
- We read 50th sample into memory
- We modify it
- We write 50th sample to disk
- We read 100th sample into memory
- We modify it
- We write 100th sample to disk...
..

Total: 20*2 requests, 4B of memory used. Was originally used for mmap path.


Actually from python parameter is used only in [reading](https://github.com/equinor/segyio/blob/0cb2a540e1e00ec92bd16b9e88d8035f81335251/lib/src/segy.c#L2187) algorithm now, but I decided it is better users have access anyway.


Let me know if you want to modify it somehow, because for me lightweight information is enough - users know better what is more important for them - requests or memory.

python/segyio/open.py
achaikou: Hm... Interesting point.

What should it really be if data already is in memory (memory_buffer), we just would be using that memory.
Is it from, in or into?

I now think that maybe function better be named `open_from_memory` ?

python/segyio/open.py
achaikou: `memory_buffer` parameter is already loaded file.
If it doesn't fit into memory user will hopefully see that before starting to work with segyio.

python/segyio/utils.py
achaikou: I think this error message is typical in the sense of shortness, but I updated it a bit and duplicated the check closer to the source. Your suggestion sounds confusing to me.

If you skip it, you get pointless error message because C-python won't know it is None.

---
We probably shouldn't check for existence of methods as it is python. You find out if something is a duck only if it quacks...
Python itself doesn't define what "file-like-object" must look like, so probably we shouldn't either. On first time use user will see...

.github/workflows/build.yaml
achaikou: `nativememory` is for memory_buffer datasource, not for stream datasource.

memory_buffer tests are run together with all standard tests, so they don't need special treatment.

`nativememory` though is planned to be used in benchmark tests, hence the option exists.

.github/workflows/build.yaml
achaikou: > My impression is that the DBUILD_PYTHON=ON is/was introduced especially for the developers who build the library on their machines via CMake and then run the tests via ctest often.

My impression too.
I used `-DBUILD_PYTHON=ON` in large tests to test this flow implicitly there (as it is not tested anywhere else).

ctest include all default python tests + c tests.
pytest include only default python ones. Or, in this case, special ones.

---

Interesting point with including it into ctests! I didn't think about it.

For the most of the development I was using a different file then and tests were slooow. But on tiny.sgy they actually can be included into developer's workflow.

Question is whether we should do it. I tested it, and on my machine `ctest` execution time went from 5.5s - 6s to 7.5s.

Do you think 25%-30% increase time is worth it, or should we rely on CI only?
CI only seems safe enough - I think datasource functionality is isolated/integrated enough that it won't be broken by unrelated development without also breaking usual ctests.

python/setup.cfg
achaikou: They are excluded because they require "datasource" parameter to be set. There is no "default" on purpose - if you are testing stream you are supposed to explicitly give me that stream you want to test.

python/test/conftest.py
achaikou: I believe we do not. My impression was that we only somewhat care about nice grouping (segyio imports near other segyio imports, etc)

python/test/conftest.py
achaikou: I also prefer pathlib, but I believe I was trying both pathlib and path here and something somewhere was ugly with pathlib and easy with path :smile: (it shouldn't be though, so probably I did something wrong).

This something might be long gone, but `os.path` stayed and I am not that inclined to check if it could be updated to pathlib :smile:

python/test/conftest.py
achaikou: Good point, but I think I would prefer to have "fixture" here, because this method actually makes a fixture we care about in tests. So "make me a fixture that makes streams"...

python/test/memory_buffer.py
achaikou: I had bugs related to opening closed file, it probably stems from this.
But here your suggestion should be fine :+1:

python/test/memory_buffer.py
achaikou: I often get confused by it myself, but I am afraid it does make ~some~ sense  :laughing:

python/test/memory_buffer.py
achaikou: `close` is called automatically on everything that is opened with `with` statement

Yes, I had problems with resource management in tests when `close()` would close all the resources and tests were configured incorrectly and would not open stuff again.

python/test/segyio_c.py
achaikou: I solemnly swear in the docs that

> Stream's close() will be called when SegyFile is closed.

In this case `fd.close()` should close the stream

python/test/segyio_c.py
achaikou: We hope it would be garbage collected, but we have no way of ensuring it.
But if we still have reference to data, it definitely won't be garbage collected. So we need to delete it first and pray to the garbage collector gods.

python/test/segyio_c.py
achaikou: Works for me every time, so I haven't thought about it....

But given that buffer is 400 bytes and I set 250 zeros 100 times, I find situation you describe unlikely. Updated 250 to 2500, if it would be good enough?

python/test/segyio_c.py
achaikou: Note that I don't do here `pytest.raises`. I think for xfail you can check only the error type, not the message. And I don't really care what type this error has, if it is wrong or not.

I struggled for a time with a bug that if exception is caught, everything exits nicely. But when exception is allowed to fly over uncaught, user gets segfault. Was related to name of the test - "memory_management_on_uncaught_error" - resource closing order and stuff.

So I want exception to fly of this test, whatever exception. (some `segy.py` tests probably assure what exceptions actually must be).

python/test/stream.py
achaikou: I think we don't :smile:

python/test/stream.py
achaikou: `small` is a standard name used for our standard testfile "small".

Used test files, at some point, were "small", "tiny" and customizable. So "testfile" is all it deserves after irritating me that much :smile:

python/test/stream.py
achaikou: It is a

> magic number that strems from how the file tiny.sgy is set up

https://github.com/equinor/segyio/blob/0cb2a540e1e00ec92bd16b9e88d8035f81335251/python/examples/make-file.py#L48

python/test/stream.py
achaikou: Because of the used assertions (see `expected` variable)

python/test/stream.py
achaikou: No for the same reason - it is not point of this test to check what exception is thrown, as long as it is thrown.

test-data/README.md
achaikou: Specific numbers depend on your machine and your network connection and actually used cloud, but I can add them to the commit message on rebase if you prefer.

But in the next commit I introduce notebook where I state that running all the tests takes me around 20s on my machine on `tiny.sgy` file on Azure datasource.

Well, `small.sgy` is around 50 times larger...

python/tutorials/using_own_stream.ipynb
achaikou: You can see what it looks like on my machine :cat:

python/tutorials/using_own_stream.ipynb
achaikou: Tried explaining it better on example, as I also don't know what properties it must fulfill because solution is unique for each stream.

python/tutorials/using_own_stream.ipynb
achaikou: Good point. Reporting numbers for local file.

(Also remember that all these numbers are very much relative. Someone with good machine and fast network connection will be way better numbers than I do).

achaikou (commented):
Thanks for the fast review! :cat: 
Some few things are updated, some are waiting your opinion :)

python/segyio/create.py
ajaust: Thanks for the explanation. This is now more clear to me. I agree that the lightweight information should be enough.

I personally would like to align the true and false explanations a bit though to use more similar terminology, i.e., use either `unnecessary data` or `memory usage` in both statements. I believe this makes it easier to understand for users with less technical intuition than us. I imagine something like

```text
minimize_requests_number : bool
        Configuration for internal algorithms. True to minimize number of
        requests to the stream at the cost of higher memory usage. False
        to minimize memory usage at the cost of more requests to the stream.
```

or 

```text
minimize_requests_number : bool
        Configuration for internal algorithms. True to minimize number of
        requests to the stream at the cost of reading/writing unnecessary data.
        False to minimize reading/writing of unnecessary data at the cost of
        more requests to the stream.
```

What do you think? Feel free to adapt to your liking.

python/segyio/open.py
ajaust: `open_from_memory` sounds good. It is also consistent with the docstring.

.github/workflows/build.yaml
ajaust: Thanks for the explanation. If you think CI only is safe enough, let's stick to that. We can integrate the tests into CTest if the code turns out to break often.

python/test/segyio_c.py
ajaust: If it is relevant/important that the data is collected, we could try using [`gc.collect`](https://docs.python.org/3/library/gc.html#gc.collect). However, I have no experience with it in terms how reliable it is as its documentation already mentions exceptions of data types that might live longer than expected (`float`).

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
...
+
+
+def test_memory_buffer_memory_management():
+    with open(testdata / "small.sgy", "rb") as f:
+        data = bytearray(f.read())
+    fd = _segyio.segyfd(memory_buffer=data, endianness=0).segyopen()
+    # assure test code owns no "data" reference, so it can be garbage collected
+    del data
+
+    # trying our best to accidently overwrite memory actually used by "data"
```
ajaust: accidently -> accidentally

python/test/segyio_c.py
```diff
@@ -592,3 +592,65 @@ def test_fread_trace0_for_depth():
...
+    strict=True
+)
+def test_memory_buffer_memory_management_on_uncaught_error():
+    with open(testdata / "small.sgy", "rb") as f:
+        data = bytearray(f.read())
+    fd = _segyio.segyfd(memory_buffer=data, endianness=0).segyopen()
+    # assure test code owns no "data" reference, so it can be garbage collected
+    del data
+
+    # trying our best to accidently overwrite memory actually used by "data"
```
ajaust: accidently -> accidentally

python/test/segyio_c.py
ajaust: If 2500 does not take tremendously longer, it is find. Otherwise, if you say 250 is enough, I trust you with that as well. I guess it is hard to guarantee anything without allocating all available memory.

python/test/stream.py
ajaust: Does this mean you would prefer to rename 

```python
small = str(testdata / "small.sgy")
```

to

```python
testfile = str(testdata / "small.sgy")
```

in `python/test/memory_buffer.py` or should it stay as it is to be consistent with already existing tests?

python/tutorials/using_own_stream.ipynb
ajaust: Awesome!

python/tutorials/using_own_stream.ipynb
ajaust: I ran the test on my machine now and it worked out nicely. 👍

ajaust (commented):
Great! I have only minor comments/answers to the open discussions and found a typo.

python/test/segyio_c.py
achaikou: It won't hurt.

python/test/segyio_c.py
achaikou: 2500 seems fine.

python/test/stream.py
achaikou: No, I don't really care what these variables are named. I do not see any significant inconsistency there as stream and memory_buffer tests are related to each other, but do not implement any "pattern".

`small` is a specific file located in testdata. It has a predefined path. Name is used in several places, including `__init__.py`
`testfile` in this case is just a name of the file. I have no clue where it is located, it is up to stream tests to figure it out.

So I still do not see any reason to update anything here.

ajaust (approved):
Awesome. Go ahead and squash the fixups and merge. 😄

PR #633, opened 2025-07-07
Python tests on increment.sgy and decrement.sgy

achaikou:


python/examples/make-field-limits.py
```diff
@@ -0,0 +1,132 @@
+import sys
+import numpy as np
+import segyio
+
+# Purpose is to check that all fields behave correctly when sign bit is set.
+# Values in the created file mostly follow certain logic.
+
```
ajaust: Should we shortly describe here what the intended output is? I was thinking about a longer version of the description in the README explaining what values are written in which field of the SEG-Y file. I think this would make it easier for "outsiders" (or our future selves) to understand what is going on without studying the code.

python/examples/make-field-limits.py
```diff
@@ -0,0 +1,132 @@
+import sys
+import numpy as np
+import segyio
+
+# Purpose is to check that all fields behave correctly when sign bit is set.
```
ajaust: What "sign bit" is meant here? There is no variable called `sign_bit` in the script and the SEG-Y specification does not refer to "setting the sign bit".

applications/test/catr-dec.output
ajaust, line 1: If I understand it correctly, this output is the trace header of the second trace. Is this correct? Should the information about which trace the header stems from be part of the file name (something like `catr-t2-dec.output`, `catr-trace-2-dec.output` or similar)? At the moment I find the information to be somewhat hidden.

applications/test/catr-inc.output
ajaust, line 1: If `applications/test/catr-dec.output` is renamed w.r.t. to trace number, this. should be renamed as well.

python/examples/make-field-limits.py
```diff
@@ -0,0 +1,132 @@
...
+            return 5
+        if field == segyio.BinField.ExtSamples:
+            return samples
+        if field == segyio.BinField.ExtendedHeaders:
+            return 0
+
+        if field == segyio.BinField.ExtInterval:
+            return 1125899906842594.0 * value_sign
+        if field == segyio.BinField.ExtIntervalOriginal:
+            return 1125899906842593.0 * value_sign
```
ajaust, line 60: How did you choose the values `1125899906842594.0` and `1125899906842593.`? Are they simply far enough from all other values?

python/examples/make-field-limits.py
```diff
@@ -0,0 +1,132 @@
...
+                base = 0
+            else:
+                # in 2-complements notation lowest value is binary 1000...0000 = 0x8000...0000
+                base = 2 ** (field_size * 8 - 1)
+        else:
+            if field in unsigned:
+                # all bits can be used to represent number
+                base = 2 ** (field_size * 8) - 1
+
+                # note: logic slip in the current implementation
```
ajaust, line 88: What do you mean by "logic slip" here?

python/examples/make-field-limits.py
```diff
@@ -92,7 +92,7 @@ def calculate_value(field, field_size, diff):
                 # highest bit is reserved for sign
                 base = 2 ** (field_size * 8 - 1) - 1
 
-        return base + diff * diff_sign
+        return base * value_sign + diff * diff_sign
```
ajaust, line 102: Good catch. 😄

python/segyio/segyio.cpp
```diff
@@ -1744,38 +1744,69 @@ PyObject* putfield( PyObject*, PyObject *args ) {
 
     switch ( fd.datatype ) {
         case SEGY_UNSIGNED_INTEGER_8_BYTE:
-            fd.value.u64 = PyLong_AsUnsignedLongLong(value_arg);
+            {
+                unsigned long long val = PyLong_AsUnsignedLongLong( value_arg );
+                if (PyErr_Occurred() || val > UINT64_MAX) {
```
ajaust, line 1757: I think we can remove the ` || val > UINT64_MAX` part from the expression

```suggestion
                if (PyErr_Occurred()) {
```

I think that the expression `val > UINT64_MAX` is always `true`. If `unsigned long long` is an unsigned 64-byte type, the maximum value it can hold is `2^64 - 1`. This is the same value that `UINT64_MAX` holds. Any larger value would (potentially silently) wrap around become smaller than `UINT64_MAX`. According to [the documentation](https://docs.python.org/3/c-api/long.html#c.PyLong_AsUnsignedLongLong) will raise an exception if the value in `value_arg` is too large for `unsigned long long` so we should be safe after all.

python/segyio/segyio.cpp
```diff
@@ -1744,38 +1744,69 @@ PyObject* putfield( PyObject*, PyObject *args ) {
...
         case SEGY_UNSIGNED_INTEGER_4_BYTE:
-            fd.value.u32 = PyLong_AsUnsignedLong(value_arg);
+            // currently no fields
+            // fd.value.u32 = PyLong_AsUnsignedLong(value_arg);
             break;
         case SEGY_UNSIGNED_SHORT_2_BYTE:
-            fd.value.u16 = PyLong_AsUnsignedLong(value_arg);
+            {
+                unsigned long val = PyLong_AsUnsignedLong( value_arg );
+                if( PyErr_Occurred() || val > UINT16_MAX ) {
```
ajaust, line 1775: If we care about C++ness of the code, we could use `[std::numeric_limits](https://en.cppreference.com/w/cpp/types/numeric_limits.html)` in for the maximum values of `UINT16_MAX`, `UINT8_MAX` etc. However, now that I have written this, I am not sure if this is worth it and if it makes anything look nicer either.

python/segyio/segyio.cpp
```diff
@@ -1744,38 +1744,69 @@ PyObject* putfield( PyObject*, PyObject *args ) {
...
         case SEGY_SIGNED_INTEGER_8_BYTE:
-            fd.value.i64 = PyLong_AsLongLong(value_arg);
+            // currently no fields
+            // fd.value.i64 = PyLong_AsLongLong(value_arg);
             break;
         case SEGY_SIGNED_INTEGER_4_BYTE:
-            fd.value.i32 = PyLong_AsLong(value_arg);
+            {
+                long val = PyLong_AsLong( value_arg );
+                if( PyErr_Occurred() || val > INT32_MAX || val < INT32_MIN ) {
```
ajaust, line 1803: I think we have a similar issue here with `val > INT32_MAX || val < INT32_MIN` as in the `SEGY_UNSIGNED_INTEGER_8_BYTE` case. `long` is  not guaranteed to have more than 32 bit. A 32-bit integer itself cannot be out of range, i.e., it cannot be larger than `INT32_MAX` or smaller than `INT32_MIN` so this part of expression is always true.

We can either rely on Python raising and error when the value is out of range or we can read the value as `PyLong_AsLongLong`, do the check and then cast it down to `long` in the assignment in line 1788.

python/segyio/segyio.cpp
```diff
@@ -1744,38 +1744,69 @@ PyObject* putfield( PyObject*, PyObject *args ) {
...
+                fd.value.i16 = val;
+            }
             break;
 
         case SEGY_IEEE_FLOAT_8_BYTE:
-            fd.value.f64 = PyFloat_AsDouble(value_arg);
-            break;
-        case SEGY_IEEE_FLOAT_4_BYTE:
-            fd.value.f32 = PyFloat_AsDouble(value_arg);
-            break;
```
ajaust: This commit removes the assignments for `SEGY_SIGNED_CHAR_1_BYTE` and `SEGY_IEEE_FLOAT_4_BYTE`, but there is not explanation for this in the commit message. 

Due to this removal the field `f32` seems to not be used anywhere in the code and could be removed from the union `segy_field_value` in `segy.h`.

The cases `SEGY_SIGNED_CHAR_1_BYTE` is still available in the `getfield` function for accessing data. Is the risk that we would access a field that is never assigned/initialised? Should we remove this case from `getfield` as well?

It would be nice to state somewhere why we remove it and why we are allowed to remove it.

python/segyio/segyio.cpp
```diff
@@ -1744,38 +1744,69 @@ PyObject* putfield( PyObject*, PyObject *args ) {
...
+                    return ValueError( "Value out of range for unsigned char" );
+                }
+                fd.value.u8 = val;
+            }
             break;
 
         case SEGY_SIGNED_INTEGER_8_BYTE:
-            fd.value.i64 = PyLong_AsLongLong(value_arg);
+            // currently no fields
+            // fd.value.i64 = PyLong_AsLongLong(value_arg);
```
ajaust: The case `SEGY_SIGNED_CHAR_1_BYTE` and `SEGY_IEEE_FLOAT_4_BYTE` are completely removed as a `case`. Why do we keep the case `SEGY_SIGNED_INTEGER_8_BYTE`, but comment it out.

We keep the case `SEGY_SIGNED_INTEGER_8_BYTE` in `getfield` (line 1697). Do we need this and is there the risk that the value for `i64` is never initialised such that we would read from a non-initialised variable?

python/test/segy.py
```diff
@@ -937,6 +937,108 @@ def test_write_header_update_atomic(small):
...
 
+def test_field_types_read():
+    with segyio.open(testdata / 'decrement.sgy', "r", ignore_geometry=True) as f:
+        bdec = f.bin
+        tdec = f.header[3]
+
+        assert bdec[BinField.NrTracesInStream] == 18446744073709551572
+        assert bdec[BinField.JobID] == 2147483647
+        assert bdec[BinField.SamplesOriginal] == 65526
+        assert bdec[BinField.VerticalSum] == 32755
```
ajaust: This might be something I have missed in the first commit, but my expectation was that I would (at least) see the maximum valid values of each field type here once, i.e. `INT32_MAX`, `UINT64_MAX` etc. This is the case for `JobID` (`2**31 - 1 =2 147 483 647 = UINT32_MAX`), but not for the other fields. I expected `18446744073709551615` for `NrTracesInStream`, `65536` for `SamplesOriginal`. 

What am I missing here? Are we not checking the first field of each available data type in the header ? If not, how did we select the fields. that we are testing here?

python/test/segy.py
```diff
@@ -937,6 +937,108 @@ def test_write_header_update_atomic(small):
...
+    with segyio.open(testdata / 'increment.sgy', "r", ignore_geometry=True) as f:
+        binc = f.bin
+        tinc = f.header[3]
+
+        assert binc[BinField.NrTracesInStream] == 43
+        assert binc[BinField.JobID] == -2147483648
+        assert binc[BinField.SamplesOriginal] == 8
+        assert binc[BinField.VerticalSum] == -32756
+        assert binc[BinField.SEGYRevisionMinor] == 37
+        assert binc[BinField.ExtIntervalOriginal] == -1125899906842593.0
```
ajaust: Basically the same comment as for `decrement.sgy` just that I would have expected to see the lower bound of all field types.

python/test/segy.py
```diff
@@ -937,6 +937,108 @@ def test_write_header_update_atomic(small):
...
+            BinField.VerticalSum: -32768,
+            BinField.SEGYRevisionMinor: 255,
+            BinField.ExtIntervalOriginal: 10.125,
+        }
+
+        tupd = {
+            TraceField.INLINE_3D: -2147483648,
+            TraceField.TRACE_SAMPLE_COUNT: 65535,
+            TraceField.DelayRecordingTime: 32767,
+        }
```
ajaust: I would consider moving the definition of `tupd` and `bupd` of the `with ... segyio.open ... as f` scope (just in front to it). `tupd` and `bupd` are used in this and the next `with ... segyio.open ... as f` scope. I was first confused where `tupd` and `bupd` came from in the second scope that compares the headers and checks that the values were written. 

Misunderstanding this might be my mistake though. I am too used to C++ in which variables within the scope would be invalidated outside of it.

python/examples/make-field-limits.py
achaikou: I was doing something else yesterday and suddenly noticed this script failed to produce what it has to. Was very confused for probably one hour because it was working before!!! 

So looks like I removed this sign right before making a PR to check that script produces different file here and then forgot to put it back :smile:

python/examples/make-field-limits.py
achaikou: The [standard](https://en.wikipedia.org/wiki/Sign_bit) one.

Most interesting one when you have to distinguish between signed/unsigned integers. That's why test files go to the extremes with their values.

python/examples/make-field-limits.py
achaikou: I am afraid describing the logic here in text can be more convoluted than simply leaving it to the code. I struggled to explain it, so didn't bother mentioning it anywhere.

So I  wrote something like that below. Do you think it would help you or confuse you even more? I don't think it will help me.

> As a general rule we iterate through binary/trace integer fields starting from the first one. For each field one can define its "difference". For nth field difference is n * step, where step is positive for increment / negative for decrement and its absolute value is 1 for binary header and 1st trace header, 2 for 2nd trace header, 3 for 3rd, etc.
> The final value of the field is its minimum (for increment) /maximum (for decrement) value for its type plus "difference".
> Thus usually neighboring fields differ from each other by 1 (2, 3, etc in case of 2-nd or further trace header) in the least significant byte.
> Some integer field values are excepted from this logic to make file correctly readable. Also fields that are not integers follow its own logic.

python/examples/make-field-limits.py
achaikou: I didn't choose them. Remember that here I add logic for creating already existing file.

I am not sure what Yngve's logic here has been, but for me interesting float values are those with something present after the dot. So I want to modify them in the future.

python/examples/make-field-limits.py
achaikou: I have to do additional "minus 1" here, for some reason. And it is not even all unsigned values, just their subset ("inconsistent").
I suspect Yngve changed the logic here at some moment, but forgot to update the file.

Fixing this slip is also in my plans for modifying this file.

applications/test/catr-dec.output
achaikou: This is correct.

I follow the existing logic where names are short and very vague.
I am sure `segyio-catr -r 4 9 -t 12 ${small} > catr.out` definitely does something more, but it is still named `catr.out`.

So per this logic it looks like suffixes inc/dec are the only ones that matter...

applications/test/catr-inc.output
achaikou: For the same reason - generally not renaming, but if you really feel that we should start breaking current pattern - will do :cat:

python/segyio/segyio.cpp
achaikou: It was added on purpose. I think we should keep it to be aligned with other unsigned checks.

Additionally:
> If unsigned long long is an unsigned 64-byte type

Apparently according to C99 standard it is _at least_ 64-bit. Obviously more is not even remotely likely, yet still.

Both together I believe justify the max border check.

python/segyio/segyio.cpp
achaikou: Damn, I always forget I am in C++ code :sweat_smile: 

From the readability point of view `UINT16_MAX` probably looks nicer? So unless you want it the other way around, I am inclined to leave it as is.

python/segyio/segyio.cpp
achaikou: There are no signed chars or 4-byte floats in specification anywhere (unless you count newly discovered mapping, but it happened after I wrote this code and we are not yet sure how we will handle that)

Signed integer 8 byte though appears in trace header extension.

---
We have a mess in getfield and in other functions because there is no logic to this (there are some going over all types, some only over specification-present ones). I actually planned to clean things up in this PR, but then we discovered trace header mapping and I decided it must be postponed until we know what to do with that.

Regarding the risk of reading uninitiated value: it is always there because we use union. We always must be sure that what we wrote and what we read is the same. So as long as all switch-cases have "defaults" that throw, we should be able to catch all inconsistencies.

python/segyio/segyio.cpp
achaikou: I do not follow.

long in python definitely can be over 32 bits.
long in C/c++ can be 64 bit (Linux) or 32 bit (Windows).
So if we are on Linux, already `INT32_MAX + 1` would only be caught by this explicit check.

python/segyio/segyio.cpp
achaikou: Explained in the other comment.
Yngve originally added them because it was easy. Now as I am adding additional checks it is easier to remove them, at least for me :smile: 

I you would feel it is safer, I can simply put it back and add the checks, even though fields are not used. And I can deal with it all together later on cleanup after we solved the mapping problem.

python/test/segy.py
achaikou: Yes, you misunderstood the logic of the file creations. Max-max value is present only in the first field, JobID. It is not that "the first field of each type has max value", it is "the first field has max value". So if the second field after JobID was uint8, it wouldn't have 255, it would already be 254. If it was third field, it would be 253, not 255. And so on.

For this test: so here I chose one value of each type. Any value. All values are close enough to maximum, so most significant byte is the same in all the values. Thus is doesn't matter what we choose for checking.

python/test/segy.py
achaikou: Same answer :smile:

achaikou (commented):
Thanks for the review!
Just one change as some other ones I prefer to confirm with you first :)

python/examples/make-field-limits.py
ajaust: I guess I got confused by the phrase "when sign bit is set". In my understanding it is always set, either to `0` or to `1` and we do not test one of the two cases, i.e., either only positive or only negative numbers.

From my understanding, we test the case of lower (sign bit is 1 = negative values) and upper range (sign bit is 0 = positive values) of allowed values.

Maybe something like this:

> Purpose is to check that all fields behave correctly if set to values from the range of largest, positive values (sign bit is `0`) or of smallest, negative for signed types, values (sign bit is `1`.

We can also leave out the reference to the "sign bit" because we do not really use it explicitly anywhere in the code so it might be more distracting to mention it.

python/examples/make-field-limits.py
ajaust: Yes, I think this might be more confusing indeed. Maybe something in between which explains the intent a little bit more, but without too much detail. Some suggestions:

> Values in all integer fields are set using increasing or decreasing numbers respecting the bounds of the field type (32-bit signed integer, 32-bit unsigned integer...). The first field is using the smallest allowed value of its type, if numbers are increased, or the largest allowed value, if numbers are decreased. For the details, please refer to the actual code.

applications/test/catr-dec.output
ajaust: Then let's keep the names the way they are. 🙂

python/segyio/segyio.cpp
ajaust: Ah, I agree. We need/expect a 8-byte integer due to the SEG-Y specification the check makes sense in case our system has, for whatever reasons, more than 64-bit. I did not think about this.

python/segyio/segyio.cpp
ajaust: For me this was a similar case as in `SEGY_UNSIGNED_INTEGER_8_BYTE`, I missed/forgot that the SEG-Y standard prescribes a 4-byte integer here so even if our system's long is larger than 32-bit, we need to check that the value is within the SEG-Y specification.

python/segyio/segyio.cpp
ajaust: Sounds good. Maybe a short note in the commit message to mention this would be great in case we check this commit sometimes in the future and don't remember what was going on, but I leave this decision to you. 🙂

python/test/segy.py
ajaust: Thanks for the explanation. Then I just have the question/suggestion if it could make sense to order the assertions in the order the fields appear in the SEG-Y, i.e., assert on `JobID` first and second whatever the next field would be that we actually check etc..

python/test/segy.py
ajaust: Same comment. 😄

ajaust (approved):
Looks good to me. I have some minor comments with suggetions of rephrasing of some comments, but that is nothing major. You can decide whether you want to change anything there. If you have any questions, just reach out to me. 😄

achaikou (commented):
Thank you, all done :)

ajaust (approved):

PR #634, opened 2025-07-09
Migrate to scikit-build-core for building the Python extension

ajaust:
Note that the additions made to `pyproject.toml` stem from the removed `setup.py` and `setup.cfg` files. The moved content is just not highlighted in the diff generated by GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,8 +201,7 @@ The segy file object has several public attributes describing this structure:
 * `f.ext_headers`
     The number of extended textual headers
 
-If the file is opened *unstructured*, all the line properties will will be
-`None`.
+If the file is opened *unstructured*, all the line properties will be `None`.
 
 ### Modes ###
 
@@ -379,7 +378,7 @@ demos in this
 ## Reproducing the test data ##
 
 Small SEG-Y formatted files are included in the repository for test purposes.
-The data is non-sensical and made to be predictable, and it is reproducible by
+The data is nonsensical and made to be predictable, and it is reproducible by
 using segyio. The test files are located in the test-data directory. To
 reproduce the data file, build segyio and run the test program `make-file.py`,
 `make-ps-file.py`, and `make-rotated-copies.py` as such:
@@ -404,8 +403,8 @@ segyread tape=small.sgy ns=50 remap=tracr,cdp byte=189l,193l conv=1 format=1 \
 suswapbytes < small.su > small-lsb.su
 ```
 
-If you have have small data files with a free license, feel free to submit it
-to the project!
+If you have small data files with a free license, feel free to submit it to the
+project!
 
 ## Examples ##
 
@@ -515,7 +514,7 @@ with segyio.open(output_file, "r+") as src:
         src.iline[i] = 2 * src.iline[i]
 ```
 
-[Make segy file from sctrach](python/examples/make-file.py)
+[Make segy file from scratch](python/examples/make-file.py)
 
 ### MATLAB ###
 


README.md
```diff
@@ -97,7 +97,7 @@ A copy of segyio is available both as pre-built binaries and source code:
 To build segyio you need:
  * A C99 compatible C compiler (tested mostly on gcc and clang)
  * A C++ compiler for the Python extension, and C++11 for the tests
- * [CMake](https://cmake.org/) version 3.11 or greater
+ * [CMake](https://cmake.org/) version 3.17 or greater
```
achaikou: Also update line 129 about PYTHON_EXECUTABLE. It should be `-DPython_ROOT_DIR` now, right?

python/setup-CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
-cmake_minimum_required(VERSION 3.11)
+cmake_minimum_required(VERSION 3.17)
 project(segyio-python-extension LANGUAGES C CXX)
 
 set(CMAKE_CXX_STANDARD 11)
 
-find_package(PythonExtensions REQUIRED)
+find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
```
achaikou: Isn't Development.Module added in version 3.18, while we migrated to 3.17?

python/setup-CMakeLists.txt
```diff
@@ -1,14 +1,13 @@
...
 find_package(segyio REQUIRED)
 
-add_library(_segyio MODULE segyio/segyio.cpp)
-python_extension_module(_segyio)
```
achaikou, line 10: This code can be gone because scikit-build-core recommends switching and because we now have this newer cmake.

Could we mention in the commit description that this is also done? Or should it go to the "replace" commit?

python/pyproject.toml
```diff
@@ -66,6 +66,7 @@ before-build = [
         -DBUILD_TESTING=OFF \
         -DBUILD_PYTHON=OFF \
         -DBUILD_BIN=OFF \
+        -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
```
achaikou, line 130: Any idea why warning
`ld: warning: object file (/usr/local/lib/libsegyio.a[2](segy.c.o)) was built for newer 'macOS' version (14.0) than being linked (11.0) `
refers to Macos 11 when version is 10.9? Is new macos simply not compatible with 10.9, so it uses first available version?

Interesting that on macos 13 message is
`ld: warning: object file (/usr/local/lib/libsegyio.a[2](segy.c.o)) was built for newer 'macOS' version (13.7) than being linked (10.9)`

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
-project(segyio-python)
+project(segyio-python LANGUAGES CXX)
 
-find_package(Python COMPONENTS Interpreter REQUIRED)
+set(CMAKE_CXX_STANDARD 11)
 
-if (PYTHON_INSTALL_LAYOUT)
-    set(setup-install-layout --install-layout ${PYTHON_INSTALL_LAYOUT})
+find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)
+# If we executate CMake via a scikit-build-core, i.e., `python build`, `pip
```
achaikou: I like this new word, "executate". It has some unique flavor to it :smile:

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
+# we can return from this file. The installation and collection of relevant
+# files is handled by scikit-build-core. Other options, like building the
+# documentation, are not available via scikit-build-core.
+if (SKBUILD)
+    return()
+endif()
+
+# Explicitly install Python files to avoid copying files that are not required,
+# e.g., segyio.cpp or folder like __pycache__ that exist if the tests were run
+# via ctest.
```
achaikou, line 71: What is the alternative?
Now every time you add new file to the library, you have to remember to put it into cmake. I would prefer copying not required files over direct listings.

I also see there is option in the docs
`install(DIRECTORY segyio DESTINATION ${SEGYIO_PYTHON_INSTALL_DIR} FILES_MATCHING PATTERN "*.py")`
Looks like it is also possible to just exclude some files.

Exclusion instead of inclusion would make more sense to me.

python/CMakeLists.txt
achaikou, line 1: Any idea why on the switch I suddenly get tons of errors for segyio.cpp, starting from
```

[ 36%] Building CXX object python/CMakeFiles/_segyio.dir/segyio/segyio.cpp.o
/home/segyio/python/segyio/segyio.cpp:1582:19: error: cast between incompatible function types from ‘PyObject* (*)({anonymous}::segyfd*)’ {aka ‘_object* (*)({anonymous}::segyfd*)’} to ‘PyCFunction’ {aka ‘_object* (*)(_object*, _object*)’} [-Werror=cast-function-type]
 1582 |     { "segyopen", (PyCFunction) fd::segyopen, METH_NOARGS, "Open file." }
```

And ending up with

`/home/segyio/python/segyio/segyio.cpp:1908:1: error: missing initializer for member ‘PyModuleDef::m_free’ [-Werror=missing-field-initializers]`

when building in Debug mode?
(due to rebase conflict, this one didn't build anything)

python/CMakeLists.txt
```diff
@@ -180,6 +188,6 @@ add_test(
 add_test(
     NAME python.example.multi-text
     COMMAND ${Python_EXECUTABLE} -m examples.make-multiple-text
-                _skbuild/multi-text.sgy
+                test-artifacts/multi-text.sgy
```
achaikou: Why is there `test-artifacts` directory in my python dir? That looks weird.

Could we instead update makefile/makepsfile/multitext to use the same solution as `python.example.write` - by writing files to `CMAKE_CURRENT_BINARY_DIR` ?

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-
-    # if DESTDIR is used, either for build and install, and make (or similar)
-    # is the generator, the scikit-build's internal install command will
-    # respect it and install the build-dir shared object there. The cmake
-    # driven install already respects DESTDIR by translating into --root, so
-    # simply remove it from then environment when doing python commands
+if (SKBUILD)
+    set(SEGYIO_PYTHON_INSTALL_DIR segyio)
+else()
```
achaikou, line 26: So if we are running without SKBUILD, we are installing into site-packages.
And if we are running with SKBUILD, we are not installing there?? Where are we installing then?

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
+    #   PREFIX  |          SITELIB
+    #
+    # This is not a path one would expect the Python library to be in.
+    set(SEGYIO_PYTHON_INSTALL_DIR lib/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages/segyio)
+endif()
 
-add_dependencies(segyio-python segyio)
-
-install(CODE "
-    if (DEFINED ENV{DESTDIR})
```
achaikou, line 87: Should Readme be updated? It still mentions some DESTDIR. Maybe something else outdated as well?..

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
+    set(SEGYIO_PYTHON_INSTALL_DIR segyio)
+else()
+    # CMake defines Python_SITELIB that points to the system path for Python
+    # libraries, but we cannot use it here. In case of a non-empty installation
+    # prefix, CMake would merge the installation path into something like
+    # ${CMAKE_INSTALL_PREFIX}/${Python_SITELIB} which would lead to the wrong
+    # full path. For example, it could look like this:
     #
-    # This is probably something that should be addressed upstream
-    COMMAND ${CMAKE_COMMAND} -E env --unset=DESTDIR
```
achaikou, line 64: Do I understand correctly that this command and install command were used before both for skbuild workflow and non-skbuild workflow?

And now they somehow both are exchanged with one-line `install` command + target property set?

I struggle to understand which line builds C-python bindings (_segyio?) both for skbuild build and for non-skbuild build.

python/CMakeLists.txt
```diff
@@ -1,93 +1,99 @@
...
-            # the build tree
-            -DCMAKE_INSTALL_RPATH_USE_LINK_PATH=ON
-            -DCMAKE_INSTALL_RPATH=$<TARGET_FILE_DIR:segyio>
-            -DCMAKE_INSTALL_NAME_DIR=$<TARGET_FILE_DIR:segyio>
-)
+    # /usr/local/usr/lib/python3.12/site-packages
+    #   PREFIX  |          SITELIB
+    #
+    # This is not a path one would expect the Python library to be in.
+    set(SEGYIO_PYTHON_INSTALL_DIR lib/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages/segyio)
```
achaikou, line 37: What command should I run to have this? I have no `segyio` directory in my site-packages.
`make install` seems to use standard C directory and I understand this is not a pip install path (as there are no examples in our Readme/CI, I assume we don't want to introduce that).

python/pyproject.toml
```diff
@@ -1,10 +1,92 @@
...
+    "Topic :: Scientific/Engineering",
+    "Topic :: Scientific/Engineering :: Physics",
+    "Topic :: Software Development :: Libraries",
+    "Topic :: Utilities",
+]
+
+
+[tool.pytest.ini_options]
+python_files = "test/*.py"
+addopts = "--ignore=test/benchmarks.py --ignore=test/large.py"
```
achaikou: This is the one in the need of updating.

python/pyproject.toml
```diff
@@ -1,10 +1,92 @@
...
+    "scikit-build-core",
 ]
+build-backend = "scikit_build_core.build"
+
+[project]
+name = "segyio"
+version = "1.9.13"
+description = "Simple & fast IO for SEG-Y files"
+authors = [
+    { name = "Equinor ASA", email = "jokva@equinor.com" },
```
achaikou: I think we really need some kind of team email. Or just leave that one empty.

python/pyproject.toml
```diff
@@ -32,8 +115,8 @@ test-requires = "pytest"
 # copy that too. segyio's python test-suite doesn't follow common naming
 # convensions for filenames. Hence pytest doesn't pick them up by default.
 # If not for powershell on windows we could simply run 'pytest test/*.py'.
-# Instead we copy setup.cfg which embeds the same pattern but doesn't rely on
-# the shell.
+# Instead we copy configure pytest in this file which embeds the same pattern
```
achaikou: What do we copy??? What is "configure pytest" ? pyproject.toml?

I also suspect this is somewhat outdated. dlisio version is somewhat more understandable, to my liking.
Also cbuildwheel swears tests are run against built package, not against local copy, so maybe these preparations are outdated.

May be something to look at now or in the future. This new `test-sources` properties from the article you linked seems relevant
```
# Copy the "tests" folder, plus "data/test-image.png" from the source folder to the test folder.
test-sources = ["tests", "data/test-image.png"]
```

python/pyproject.toml
```diff
@@ -44,10 +127,12 @@ test-command = [
...
     """,
     """python {project}/.github/utils.py copy \
-        --src {package}/setup.cfg             \
-        --dst tmptest/python/setup.cfg        \
+        --src {package}/pyproject.toml        \
+        --dst tmptest/python/pyproject.toml   \
     """,
     """cd tmptest/python""",
+    """echo \"tmptest/python\"""",
+    """ls -l""",
```
achaikou: Debug remnants or do you want to leave that?

python/pyproject.toml
```diff
@@ -66,7 +151,6 @@ before-build = [
         -DBUILD_TESTING=OFF \
         -DBUILD_PYTHON=OFF \
         -DBUILD_BIN=OFF \
-        -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 \
```
achaikou: Why did you remove it from here (macos specialized) and added to main cbuildwheel flow?

achaikou (commented):
Awesome! :heart_eyes_cat: 
I am still quite confused about the build process, so here go a few stupid questions.

And it definitely needs rebasing. I want to see what that all looks like in CI and figure out what I need to type to invoke which paths...

python/pyproject.toml
ajaust: This are debugging remnants. I will remove it.

python/CMakeLists.txt
ajaust: Yes, lets use `CMAKE_CURRENT_BINARY_DIR`.

python/pyproject.toml
ajaust: I have no proof for this, but my guess is the following:

MacOS 11 ([Big Sur](https://en.wikipedia.org/wiki/MacOS_Big_Sur)) is the first one that supports Apples Arm64 CPUs. Therefore, the oldest deployment target for the `arm64` build is `11.0`. MacOS 10.9 ([OS X Mavericks](https://en.wikipedia.org/wiki/OS_X_Mavericks)) was available for `amd64` so the CI/cibuildwheel can build with `10.9` as target.

python/CMakeLists.txt
ajaust: If we run with SKBUILD, then we install into `.../site-packages/segyio`. The path to `.../site-packages` is automatically prepended by scikit-build-learn. Basically, scikit-build-core sets the `CMAKE_INSTALL_PREFIX` to the `site-packages` directory automatically. We set the path that is appended to the prefix path here. 

Note, we could also configure the cmake prefix path via the `pyproject.toml` for builds triggered by scikit-build-core.

python/CMakeLists.txt
ajaust: In the old case, the actual Python extension would always be build by `setup-CMakeListst.txt` via scikit-build, but the path to get there depends on how the build is initially triggered.

I hope this is not completely confusing. 😅 

### Build the extension via Python/cibuildwheel

Roughly the following would happen from within the `python/` directory:

1. Run some command like `python -m build --wheel` .
2. Trigger a build via SKBUILD that parses the python config and setup files (`setup.py`, `pyproject.toml`, and `setup.cfg`).
3. SKBUILD triggers a build via CMake parsing `python/CMakeLists.txt`.
4. CMake would immediately jump into `setup-CMakeLists.txt` due to line 6.
5. The `_segyio.pythonverstionstring.so` would be build
6. CMake returns from `setup-CMakeLists.txt` into `CMakeLists.txt`
7. CMake returns due to `return ()` in line 7.
8. Done

### Build the extension via CMake

If the build was initially triggered by CMake, i.e., it was configured with `BUILD_PYTHON=ON`, and the user then runs `make install`.

1. Build segyio core library.
2. Eventually, enter the `python/CMakeLists.txt`. The build was **not** triggered by SKBUILD so we skip lines 1-8. 
3. From inside `python/CMakeLists.txt` call `python setup.py build_ext ...`, the command specified here in line 44. This triggers building the extension via SKBUILD. 
4. The steps 2-8 from above under "Build the extension via Python/cibuildwheel" because we are in the "Python/cibuildwheel build path" now.
5. Build Python docs (optional)
6. Add Python test to CTest testsuite.
7. Done

python/CMakeLists.txt
ajaust: Let's assume you have Python 3.12 and you build via CMake with `BUILD_PYTHON=ON` and the `CMAKE_PREFIX_PATH=/tmp/usr/local` (that is what I use for test). When running `make install` the Python extension should be installed to `/tmp/usr/local/lib/python3.12/site-packages/segyio`. This is the same behavior the build system via scikit-build (non-core) has.

Now, this line ensures that `python3.12/site-packages/segyio` is appended to the `CMAKE_INSTALL_PREFIX`.

python/CMakeLists.txt
ajaust: Do you trigger the build via Python (`python -m build --wheel`) or directory via CMake?

No, I have not seen this error yet. You could try to set `project(segyio-python LANGUAGES CXX)` to `project(segyio-python LANGUAGES)` again and see if this make a difference.

I also had some problems when system path was pointing to a `libsegyio.a` that was from an incompatible branch and I had forgotten to update the path.

python/CMakeLists.txt
ajaust: I think I have to look into this if we want to keep mentioning `DESTDIR`. `DESTDIR` iteslf is a [feature of make](https://www.gnu.org/prep/standards/html_node/DESTDIR.html). It is automatically supported when running make, but I see that with my changes it will install into the wrong directory, namely `$DESTDIR/$CMAKE_INSTALL_PREFIX`, for a CMake build:

```bash
$ make DESTDIR=/tmp/usr/local2/ install
[  0%] Built target doc
[  6%] Built target segyio
[ 20%] Built target c.segy
[ 26%] Built target c-include.segy
[ 33%] Built target cpp-include.segy
[ 40%] Built target _segyio
[ 46%] Built target apputils
[ 53%] Built target segyinfo
[ 60%] Built target segyinspect
[ 66%] Built target segyio-cath
[ 73%] Built target segyio-catb
[ 80%] Built target segyio-catr
[ 86%] Built target segyio-crop
[ 93%] Built target flip-endianness
[100%] Built target test-app-output
Install the project...
-- Install configuration: ""
-- Installing: /tmp/usr/local2/tmp/usr/local/lib/libsegyio.a
-- Installing: /tmp/usr/local2/tmp/usr/local/include
-- Installing: /tmp/usr/local2/tmp/usr/local/include/segyio
-- Installing: /tmp/usr/local2/tmp/usr/local/include/segyio/segy.h
-- Installing: /tmp/usr/local2/tmp/usr/local/share/segyio/cmake/segyio-config.cmake
-- Installing: /tmp/usr/local2/tmp/usr/local/share/segyio/cmake/segyio-config-noconfig.cmake
....

File: python/CMakeLists.txt (hunk 6 above)
Comment (ajaust) [reply python/CMakeLists.txt:71]: I will have a look. I guess I am under heavy impression of the [advise to not use globbing](https://cmake.org/cmake/help/latest/command/file.html#filesystem) in CMake that I always tend to list source files explicitly just to be safe.

File: python/pyproject.toml (hunk 13 above)
Comment (ajaust) [reply python/pyproject.toml]: I will try to remove the email and see if scikit-build-core is ok with that.

File: python/pyproject.toml (hunk 12 above)
Comment (ajaust) [reply python/pyproject.toml]: Do you mean adding ` --ignore=test/stream.py` (like on current main)?

File: python/pyproject.toml (hunk 14 above)
Comment (ajaust) [reply python/pyproject.toml]: I will rephrase that. I don't know how this ended up being such gibberish.

I can give it a quick try if this works nicely with the `test-sources` parameter.

File: python/pyproject.toml (hunk 16 above)
Comment (ajaust) [reply python/pyproject.toml]: Good catch. I think I messed up when I did a cherry pick from another branch. It shoud stey here in the `tool.cibuildwheel.macos` section and not appear in the `tool.cibuildwheel` section.

File: python/CMakeLists.txt
Comment (achaikou) [reply python/CMakeLists.txt:1]:
```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_C_FLAGS=\"-Wextra -Wall\" -DBUILD_BIN=ON -DCMAKE_CXX_FLAGS=\"-Wextra -Wall\""
make
```

Running on skbuild: passes
Running on skbuild-core: fails

Release builds pass.
`project(segyio-python LANGUAGES CXX)` is already there so not sure what you mean.

Does that mean that before some of my commands were not passed to the build system and now they are?

File: python/CMakeLists.txt (hunk 10 above)
Comment (achaikou) [reply python/CMakeLists.txt:64]: Thanks for explanation. :heart:
I was always confused what was going on there, but hopefully will understand enough for the purpose of this review :smile: 

So now "Build the extension via Cmake" doesn't even touch skbuild-core? How is it built now, if we removed  `COMMAND ${CMAKE_COMMAND} -E env --unset=DESTDIR ${PYTHON_EXECUTABLE} ${setup.py}` from cmake path? Or is skbuild not actually required to build extension itself, but it just knows how to connect the wheel to C library?

I doubt that I understand skbuilds purpose correctly anymore.

File: python/CMakeLists.txt (hunk 6 above)
Comment (achaikou) [reply python/CMakeLists.txt:71]: I was looking at https://cmake.org/cmake/help/v3.17/command/install.html
Seems like no globbing there

File: python/pyproject.toml (hunk 12 above)
Comment (achaikou) [reply python/pyproject.toml]: Yes. Rebasing so I would be able to see tests outputs would be great :)

File: python/CMakeLists.txt (hunk 11 above)
Comment (achaikou) [reply python/CMakeLists.txt:37]: Thanks, I was able to build it both ways now!
I didn't bother installing segyio locally before, so didn't know it installs into `/tmp/usr/local/lib/python3.12/site-packages/segyio` -kind path.

File: python/setup-CMakeLists.txt (hunk 2 above)
Comment (ajaust) [reply python/setup-CMakeLists.txt]: 1/2: I see that I did not comment on this so I will do this now here:

Good catch. Do you have a preference whether we should bump the CMake requirement to 3.18 or simply change this to `Development`? Just changing it to `Development` would be less work, but as we are already touching the CMake version here it would be simple to just bump it to 3.18.

I would currently prefer bumping CMake to 3.18 such that we can use `Development.Module`. 3.18 (July 2020) and and 3.17 (March 2020) were release quite close to each other so I don't expect that 3.18 would give more compatibility issues with old platforms than 3.17. Then we can also use the most "modern" way of integrating the Python build process.

File: python/setup-CMakeLists.txt (hunk 3 above)
Comment (ajaust) [reply python/setup-CMakeLists.txt:10]: 2/2: I see that I did not comment on this so I will do this now here:

It will be a bit more work, but I think I would prefer to move it to the `moving to scikit-build-core` commit. I was not aware that `PythonExtensions` is a part of `scikit-build`, but assumed that is the old CMake way of interacting with Python. That is why I changed it as part of this commit instead of the commit where we move to `scikit-build-core`.

File: python/setup-CMakeLists.txt (hunk 2 above)
Comment (achaikou) [reply python/setup-CMakeLists.txt]: No preference! I have no idea what would be the difference between Development and Development.Module, so if you prefer upgrading to 3.18, I am for it too :)

File: python/CMakeLists.txt
Comment (ajaust) [reply python/CMakeLists.txt:1]: I could reproduce this with `scikit-build` on the main branch. I think, `scikit-build` simply does not forward the `DCMAKE_CXX_FLAGS ` and `DCMAKE_C_FLAGS ` to the Python build process.

Short version (since GitHub deleted my long answer):

## Steps to reproduce

It is crucial to delete `python/_skbuild/`. It contains some Python build artifacts. If these are present and one changes some compile options, the build process may find the existing files and skip rebuilding them from source.

```bash
rm -rf python/_skbuild/
```

Then build as you did, but specify, e.g., the `CXXFLAGS` during the build step:

```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_C_FLAGS=\"-Wextra -Wall\" -DBUILD_BIN=ON -DCMAKE_CXX_FLAGS=\"-Wextra -Wall\""
CXXFLAGS="-Wextra -Wall" make
```

Then I see the warnings:

```text
[...]
-- Found PythonLibs: /opt/homebrew/opt/python@3.12/Frameworks/Python.framework/Versions/3.12/lib/libpython3.12.dylib (found version "3.12.11")
-- Performing Test Weak Link MODULE -> SHARED (gnu_ld_ignore) - Failed
-- Performing Test Weak Link MODULE -> SHARED (osx_dynamic_lookup) - Success
_modinit_prefix:PyInit_
-- Configuring done (3.4s)
-- Generating done (0.0s)
-- Build files have been written to: /Users/AEJ/projects/github/segyio/python/_skbuild/macosx-10.9-arm64-3.12/cmake-build
[ 50%] Building CXX object CMakeFiles/_segyio.dir/segyio/segyio.cpp.o
/Users/AEJ/projects/github/segyio/python/segyio/segyio.cpp:1690:19: warning: cast from 'PyObject *(*)(segyfd *, PyObject *, PyObject *)' (aka '_object *(*)((anonymous namespace)::segyfd *, _object *, _object *)') to 'PyCFunction' (aka '_object *(*)(_object *, _object *)') converts to incompatible function type [-Wcast-function-type-mismatch]
 1690 |     { "segyopen", (PyCFunction) fd::segyopen,
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/AEJ/projects/github/segyio/python/segyio/segyio.cpp:1692:19: warning: cast from 'PyObject *(*)(segyfd *, PyObject *, PyObject *)' (aka '_object *(*)((anonymous namespace)::segyfd *, _object *, _object *)') to 'PyCFunction' (aka '_object *(*)(_object *, _object *)') converts to incompatible function type [-Wcast-function-type-mismatch]
 1692 |     { "segymake", (PyCFunction) fd::segycreate,
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]
```

I also found [this stackoverflow post](https://stackoverflow.com/questions/10264080/python-c-extension-why-are-methods-that-use-keyword-arguments-cast-to-pycfunct) mentioning that the case errors are safe to ignore. I did not fully understand it yet though.

ajaust: TL;DR I would be in favor of doing nothing/little, but if we commonly use stricter and additional warnings, then we should find a good solution to this issue.

## More extensive thoughts

I did think about it a bit more. The main cause of this behavior is the fact that `scikit-build-core` picks up CMake options when the bindings are built while `scikit-build` ignored them or only picked up few CMake options. 

The cause of this new behavior could either be that `scikit-build-core` is closer to CMake or because I removed one layer of indirection. The pure CMake build path does not call Python anymore to build the Python bindings.

In the case of the command line

```text
cmake .. -DCMAKE_BUILD_TYPE=Debug -DBUILD_SHARED_LIBS=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_C_FLAGS="-Wextra -Wall" -DBUILD_BIN=ON -DCMAKE_CXX_FLAGS="-Wextra -Wall"
```
additional warnings are enabled (`"-Wextra -Wall"`) and the warnings are made into errors (`DCMAKE_COMPILE_WARNING_AS_ERROR=ON`). We get the same errors when using `scikit-build` if we forward these options to the build process.

The main question: Do we want to do anything about this when moving to `scikit-build-core`?

Ideas for things to do:
- Document the behavior and what warning to ignore/deactivate as compile option.
- Locally deactivate the warning using preprocessor `pragma`s.
- Try to fix this behavior, but I am not sure if this is fixable or if the Python interface uses some C-specific tricks that compilers identify as being illegal.

python/CMakeLists.txt
ajaust: I have to revise my opinion about the installation behavior. From some further research it is correct that `make DESTDIR=...` will install into `$DESTID/$CMAKE_INSTALL_PREFIX`, see [GNU docs](https://www.gnu.org/software/make/manual/html_node/DESTDIR.html). This means that the observed behavior using the changes from this PR is in accordance with my expectations after reading the documentation.

I don't think we should manipulate `DESTDIR` in our build process. There is no apparent need to manipulate it when using `scikit-build-core` as the build process succeeds. 

However, if I remember correctly, the behavior changes now to the standard behavior. I have to see if/how we should document this.

The README might need some update.

python/CMakeLists.txt
ajaust: We can work around it and collect files using `file(GLOB ...)`, but in the [CMake docs' filesystem section](https://cmake.org/cmake/help/v3.17/command/file.html#filesystem) advises against it for source files

> Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

[On Stackoverflow](https://stackoverflow.com/questions/1027247/is-it-better-to-specify-source-files-with-glob-or-each-file-individually-in-cmak) this is also discussed a lot with the general tendency to list files explicitly.

I personally prefer the explicit list of files assuming that
1. The number of files does not change often.
2. The tests are good enough to detect missing files in case we add new files.

python/CMakeLists.txt
ajaust: Yes, I think you understanding (from what I read into your questions) is correct.

Building the extension, i.e., the shared library that is used from Python (called something like `_segyio.cpython-312-darwin.so`) is always done via CMake. Technically, that was the case before this PR as well. The build was described in `python/setup-CMakelists.txt`. However, the previous build process was more convoluted as it would always go through `scikit-build` at some point. I don't know why the build was always done via `scikit-build`. It could very well be that this was due to some limitations in the CMake at that time or it felt more consistent to do it that way.

With this PR the flow of the build process is simplified. The main differences is who/what is responsible for collecting relevant Python files and placing them at the correct location

1. Build via CMake, e.g., via `cmake .. -DBUILD_PYTHON=ON`: CMake compiles the extension. In this case, we need to set the full installation path (including the Python version) and tell CMake which Python files to copy as well. `scikit-build-core` is not involved here. 
  In this case, the build process is started from the root of the repository. This means the `CMakeLists.txt` from the root is parsed first which includes `CMakeLists.txt` files from several subdirectories. We can assume that the segyio library is part of this build process such that we do not have to search for the package `segyio` (no `find_package(segyio)` needed).
2. Build via Python, e.g., `python -m build --wheel` or `pip install .` from within the `python/` directory: The Python build system will parse the `pyproject.toml` and invoke `scikit-build-core` as build backend. This will trigger the build of the extension via CMake similarly to step 1. We can be less explicit about the installation path and Python files. `sciki-build-core` does some additional magic such that the Python build system can build a legal wheel with the correct file structure and metadata.
  In this case the build process is started from `python/`. This means only `python/CMakeLists.txt` is part of the build process. No other `CMakeLists.txt` file is used.

python/setup-CMakeLists.txt
ajaust: I bump the version to 3.18.

python/setup-CMakeLists.txt
ajaust: I prepare a fixup for dropping the changes here such that these skbuild specific lines are dropped in the "replace" commit. This leads to an odd situation that I have a fixup commit that is in between the already existing commits. It does not seem like I can easily add the commit at the end because the changed file does not exist anymore at the newest commit.

python/pyproject.toml
ajaust: I am amazed. `test-sources` works and looks so much simpler. 🙂 

However, the benchmark Action depends on an old version (<3.0.0) of cibuildwheel for the reference commit which does not support `test-sources`. I added a workaround for this which updates to a new cibuildwheel version for the new commit. Do think this is a problem?

ajaust (commented):
I added some comments and some commits for fixing the mentioned issues. Let's also shortly meet to see if we can clarify anything before you look into this again. 🙂

python/CMakeLists.txt
achaikou: The only reason this doesn't fail our CI is because we do not build in Debug mode with Python enabled.
Maybe we actually should.

From what I understand we can fix only initialization warnings, but fields that we need to fix are python specific. So in python 3.13 there is an argument that does not exist in python 3.9, so if we want to properly fix all warnings we would have to branch on python version.

So I think we best just locally deactivate with `cast-function-type` and `missing-field-initializers` pragmas.

python/CMakeLists.txt
achaikou: I don't understand why you keep talking about `file(GLOB ...)` thing :question:
It wasn't even my suggestion :smile: 

Yes, docs discourage you from using this `file(GLOB)`, but they [don't discourage](https://cmake.org/cmake/help/v3.18/command/install.html#installing-directories) you at all from using `install( ... PATTERN)`  that automatically collects `.py` files.

```
install(DIRECTORY segyio/ DESTINATION ${SEGYIO_PYTHON_INSTALL_DIR} FILES_MATCHING PATTERN "*.py")
install(DIRECTORY segyio/su/ DESTINATION ${SEGYIO_PYTHON_INSTALL_DIR}/su FILES_MATCHING PATTERN "*.py")
```
It is highly unlikely that any `.py` file in `segyio` directory shouldn't be installed.
Also there is no danger of source not picked up due to "no cmake file changes". cmake-s docs [say](https://cmake.org/cmake/help/book/mastering-cmake/chapter/Install.html)

> The scripts are executed at install time to perform the actual installation of files.

- With explicit file names solution I will have to remember to add new file to the list (and spend some frustrating time thinking what I did wrong, like a few days ago when I did not add a new test file with `configure_file` and spent over an hour not having a clue what is going on).
- With pattern this is done automatically.


Could it be that your concerns regarding this are a bit outdated?
(Not arguing further though if I still did not convince you :smile: )

achaikou (commented):
I see your concernes about benchmarks.
I am not sure if this is a github-related fluctuation or an actual problem.

Main read/write tests usually are fine.
This PR struggles with at least one test, both in your fork, my fork and main repo.
And it is not good, because it gives us 20% slowness on reading (and may be writing?) large data.

I ran the commits and `Let cibuildwheel choose Linux base image` is the one where slowness appeared, then disappeared on rerun.

So either something has changed in new manylinux, or some github caching performance thing happens.
We need to look closer into this.

python/CMakeLists.txt
ajaust: Then it was a misunderstand on my side. 🙈  I think using the `FILES_MATCHING PATTERN` looks nice. 🙂  We can even get away with a single `install` statement because CMake will go through subdirectories recursively. I add a fixup.≈

python/CMakeLists.txt
ajaust: TL;DR

- I dug a bit deeper and can confirm your findings, i.e, we cannot fix all issues reported by the compiler, but most. It seems that the windows compiler does not show the error?!
- Do we have a reason why we, so far, only compile the core library with all warnings as errors (`build.yml`)? Was this done on purpose?
- Tests with all warnings as errors and additional warnings are not of the CI yet. Should we change the CI and fix the warnings in another PR or as part of the current PR?

I did play around with the fixes and additional warnings in the CI (as part of the `cibuildwheel` jobs) [in this branch](https://github.com/ajaust/segyio/pull/6/commits). [This commit](https://github.com/ajaust/segyio/pull/6/commits/f82691bc7bf3a59656ba78392440b3f99bed4036) introduces all the relevant pragmas to suppress the warnings. The commits afterwards try to fix warnings as much as possible. Commits at the end enable more warnings in the CI.

---

- I can solve `missing-field-initializers` issues, but need to branch on Python version. The branching is only needed for `PyTypeObject Segyfd`.
- I can fix most of the `cast-function-type` issues. The main issue seems to be that `PyMethodDef` expects pointers to `PyCFunction`. `PyCFunction` describes a function with two input parameters (`self` and an additional parameter). 
  - Methods without additional parameter can be rewritten to have two parameters like `PyObject* close( segyfd* self, PyObject* Py_UNUSED(args) )
`. `Py_UNUSED` is some construct to avoid compiler warnings. This is also [mentioned in the Python docs](https://docs.python.org/3/c-api/structures.html#c.METH_NOARGS).
  - Methods that have the additional parameter are fine.
  - Methods with keyword arguments have two additional parameters in total. Casting to `PyCFunction` creates some error due to that. This concerns three methods in our case:
    ```python
    { "segyopen", (PyCFunction) fd::segyopen,
      METH_VARARGS | METH_KEYWORDS, "Open file." },
    { "segymake", (PyCFunction) fd::segycreate,
      METH_VARARGS | METH_KEYWORDS, "Create file." },

    { "suopen", (PyCFunction) fd::suopen,
      METH_VARARGS | METH_KEYWORDS, "Open SU file." },
    ```
    I don't think we can fix these without disabling/accepting warnings.
- Using `pragma`s to disable warnings may interfere with the `-pedantic` compile option (used in some of the jobs in `build` but these don't build the Python bindings). The pragmas are compiler specific and thus not in the C nor C++ standard which can lead to errors with `-pedantic`. I think this mainly occurred with GCC in the CI.
- Using `pragma`s can lead to additional warnings about pragmas that the current compiler does not understand. GCC will not understand/know the clang and MSVC pragma and vice versa. Depending on the compile option this may lead to compile errors again (Using `/W4 /Wall` for windows fails the compilation due to unknown pragmas). `-pedantic` for GCC and clang seems to be fine (?!). At least there is a pragma for disabling such warnings.

- Do we want to fix the cause of warnings as much as possible or just turn of warnings where necessary?

1. Disable warnings where needed.
2. Fix as many causes of warnings and disable remaining warning for the methods with three input parameters.
3. Accept the warnings and don't compile with warnings as errors, i.e., not setting `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`. if one also compiles the Python bindings.
4. Reproduce the old build procedure that silently drops compiler flags, i.e., let `CMake` call `python build` which calls `CMake` again. 

I am definitely against Option 4.

---

I think it depends on what test strategy we want specifically for the Python bindings. That means, we have to see if we want to turn on all warnings for Python bindings as well as part of a CI job. Would we want a super strict build as part of the `build` workflows or rather as part of (all?) wheels being build?

---

Just for reference/understanding: The issue came to light due enabling more warnings and making them into errors while also building the Python bindings (-DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_C_FLAGS="-Wextra -Wall" -DCMAKE_CXX_FLAGS="-Wextra -Wall"). It is independent of the build type. You get the same errors for `Release` builds if you specify the additional CMake flags.

Another part of this issue is that the old Python build process via scikit-build would ignore the extra build flags that make errors into warnings and turns on extra warnings. This makes it look as if the issue did not appear before this PR. We can go back to doing this, i.e., calling `python build` from within CMake, but I don't think that this is a good idea.

python/pyproject.toml
```diff
@@ -117,9 +117,6 @@ test-requires = "pytest"
 test-command = [ "cd python", "pytest test" ]
 test-sources = [ "python/test", "python/pyproject.toml", "test-data" ]
 
-[tool.cibuildwheel.linux]
-manylinux-x86_64-image = "manylinux2014"
-
```
ajaust, line 58: Today the benchmarks seem to pass consistently. I don't want this change to delay the PR too much. In doubt, I would prefer to remove this commit from this PR into its own PR so we can do further investigations.

python/CMakeLists.txt
achaikou: > Do we have a reason why we, so far, only compile the core library with all warnings as errors (build.yml)? Was this done on purpose?

I think I introduced more strict compilation this year because revision 2.1 introduced different types, so this was done to catch all the "implicit conversion" things and other things we who are not very experienced in C can miss :) 


As the issues we can fix are still just subsets of those we can't and those still are just python internal workings, I think we should `just turn of warnings where necessary`. So I vote for your option 1. For me just adding
```
#if defined(__GNUC__) && !defined(__clang__)
#pragma GCC diagnostic ignored "-Wcast-function-type"
#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
#endif
```

lines at the top of `segyio/segyio.cpp` fixed all the issues. But you clearly know more about pragma hell then me, so if that is a problem, then option 3 it is.

I also think that we should not enable the check in wheels. In my mind wheels are for building the final product, so as long as tests work on most basic build, we are good.

In "build" workflow we can add as many checks as we want though.  I think adding just python debug job with `-Wextra -Wall` enabled is enough to be sure we caught all :smile: .

I am fine with taking it as a separate PR, if you want :+1:

python/pyproject.toml
achaikou: Yes, last one looks good.
So lets then assume it is a fluctuation, merge it and see how it behaves on main.

PR #635, opened 2025-07-11
Add datasource benchmark tests

achaikou:
Agreed to try running regression for all types of datasources on each PR and then see how troublesome it is.


yngve793 (approved):
Compiles and run tests locally without any problems. Nice touch to allow the user to select test options when running tests on Github. If it is possible it would be nice with a fast option, like 1-2 minutes. Fastest option I fond was 13 minutes with memory only.

achaikou:
As this is 2.0 functionality, we decided it is better to be added after 2.0 release.

PR #636, opened 2025-07-25
Segy field data cleanup

achaikou:
C get/set field cleanup that came up before and during proof-of-concept implementation for tracemaps.
Seems like nothing of the removed code is needed as it still would be us who would parse and return all fields from all trace headers.

### DIFF
--- a/applications/segyinfo.c
+++ b/applications/segyinfo.c
@@ -8,10 +8,10 @@
 
 static void printSegyTraceInfo( const char* buf ) {
     int cdp, tsf, xl, il;
-    segy_get_field_i32( buf, SEGY_TR_ENSEMBLE, &cdp );
-    segy_get_field_i32( buf, SEGY_TR_SEQ_FILE, &tsf );
-    segy_get_field_i32( buf, SEGY_TR_CROSSLINE, &xl );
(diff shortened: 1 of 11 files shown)


applications/segyinfo.c
```diff
@@ -8,10 +8,10 @@
 
 static void printSegyTraceInfo( const char* buf ) {
     int cdp, tsf, xl, il;
-    segy_get_field_i32( buf, SEGY_TR_ENSEMBLE, &cdp );
```
yngve793, line 11: Type specific functions was added by POs request.

lib/src/segy.c
```diff
@@ -833,50 +833,57 @@ int segy_init_field_data( int field, segy_field_data* fd ) {
...
+
+    int offset = field - 1;
+    if (offset >= SEGY_TEXT_HEADER_SIZE) {
+        offset -= SEGY_TEXT_HEADER_SIZE;
+    }
     uint64_t val;
     switch ( fd->datatype ) {
 
         case SEGY_SIGNED_INTEGER_8_BYTE:
-            memcpy( &(fd->value.i64), header + (fd->field_index -1), formatsize( fd->datatype ) );
```
yngve793, line 824: -1 was used to reduce the number of required variables.

PR #637, opened 2025-08-01
Improve strings handling

achaikou:
1. ASCII encoding support was added as it was requested several times already.
2. EBCDIC tables are updated according to SEG-Y specification.
3. Not yet done string-related breaking changes are applied, as well as I understood them.

Those might cause some slight breaking changes for the users, but most likely only if files were weird-ish or there was some hardcoded way to deal with ASCII files.

Note: even though strings and encodings are now supported, Trace Header bytes 233-240 are not updated to read string from it. I did that originally, but then after discussion with PO we agreed to leave Unassigned 1 and Unassigned 2 as is. Reason for this is that in revision 1.0 space is defined as
`233-240 Unassigned — For optional information`
and someone might have used that as ints before.

### DIFF
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -680,6 +680,7 @@ segy_file* segy_open( const char* path, const char* mode ) {
     // on init assume a size of 4-bytes-per-element and big-endian
     ds->elemsize = 4;
     ds->lsb = false;
+    ds->encoding = SEGY_EBCDIC;
 
     ds->minimize_requests_number = true;
     ds->memory_speedup = false;
@@ -721,6 +722,7 @@ segy_datasource* segy_memopen( unsigned char* addr, size_t size ) {
     // on init assume a size of 4-bytes-per-element and big-endian
     ds->elemsize = 4;
     ds->lsb = false;
+    ds->encoding = SEGY_EBCDIC;
 
     ds->minimize_requests_number = false;
     ds->memory_speedup = true;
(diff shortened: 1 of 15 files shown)


lib/src/segy.c
```diff
@@ -32,30 +32,30 @@
 static const unsigned char a2e[256] = {
     0,  1,  2,  3,  55, 45, 46, 47, 22, 5,  37, 11, 12, 13, 14, 15,
     16, 17, 18, 19, 60, 61, 50, 38, 24, 25, 63, 39, 28, 29, 30, 31,
-    64, 79, 127,123,91, 108,80, 125,77, 93, 92, 78, 107,96, 75, 97,
+    64, 90, 127,123,91, 108,80, 125,77, 93, 92, 78, 107,96, 75, 97,
     240,241,242,243,244,245,246,247,248,249,122,94, 76, 126,110,111,
     124,193,194,195,196,197,198,199,200,201,209,210,211,212,213,214,
-    215,216,217,226,227,228,229,230,231,232,233,74, 224,90, 95, 109,
+    215,216,217,226,227,228,229,230,231,232,233,0,  224,0,  0,  109,
```
yngve793: It is a bit strange to have a one way mapping for ASCII x5B [ writes as ¢

lib/src/segy.c
```diff
@@ -40,31 +40,31 @@ static const unsigned char a2e[256] = {
     151,152,153,162,163,164,165,166,167,168,169,192,79, 208,161,7,
     32, 33, 34, 35, 36, 21, 6,  23, 40, 41, 42, 43, 44, 9,  10, 27,
     48, 49, 26, 51, 52, 53, 54, 8,  56, 57, 58, 59, 4,  20, 62, 255,
-    65, 66, 74, 68, 69, 70, 106,72, 73, 81, 82, 83, 95, 85, 86, 87,
-    88, 89, 98, 99, 100,101,102,103,104,105,112,113,114,115,116,117,
-    118,119,120,128,138,139,140,141,142,143,144,154,155,156,157,158,
-    159,160,170,171,172,173,174,175,176,177,178,179,180,181,182,183,
-    184,185,186,187,188,189,190,191,202,203,204,205,206,207,218,219,
-    220,221,222,223,234,235,236,237,238,239,250,251,252,253,254,0
+    0,  0,  74, 0,  0,  0,  106,0,  0,  0,  0,  0,  95, 0,  0,  0,
```
yngve793: "By this change we prevent users from using characters not defined by
EBCDIC standard."
Do we need a tool to make text segyio compliant?

lib/src/segy.c
```diff
@@ -1226,6 +1228,28 @@ int segy_set_endianness( segy_datasource* ds, int endianness) {
...
+        case SEGY_EBCDIC:
+        case SEGY_ASCII:
+            break;
+        default: {
+            encoding = SEGY_EBCDIC;
+            int err = ds->seek( ds, 0, SEEK_SET );
+            if( err == 0 ) {
+                char c;
+                err = ds->read( ds, &c, 1 );
+                if( err == 0 && c == 67) {
```
yngve793, line 1243: Consider adding comment.
"If first symbol is C[Capital C]"

PR #638, opened 2025-09-03
D8 initial mapping

achaikou:


lib/src/segy.c
```diff
@@ -887,41 +887,37 @@ int segy_get_field( const char* header, int field, segy_field_data* fd ) {
     }
 }
 
-static int fd_get_int( const segy_field_data* fd, int* val ) {
-    switch( fd->datatype ) {
+int segy_get_field_int( const char* header, int field, int* val ) {
+    segy_field_data fd;
+    int err = segy_get_field(header, field, &fd);
+    if ( err != SEGY_OK ) return err;
```
ajaust, line 928: I think this could be `const int err`.

python/segyio/segyio.cpp
```diff
@@ -1710,12 +1710,22 @@ PyObject* getfield( PyObject*, PyObject *args ) {
...
+                buffer.buf<const char>(), field, &fd
+            );
+            break;
+        case SEGY_TRACE_HEADER_SIZE:
+            err = segy_get_tracefield(
+                buffer.buf<const char>(), field, &fd
+            );
+            break;
+        default:
+            return BufferError( "buffer too small" );
```
yngve793, line 1727: Buffer is of unknown size.

lib/src/segy.c
```diff
@@ -1104,12 +1094,11 @@ int segy_field_forall( segy_datasource* ds,
...
 
-        segy_field_data fd;
-        err = segy_get_field( header, field, &fd );
-        if( err != 0 ) return err;
-        err = fd_get_int( &fd, &f );
+        // note: for the moment function still works only on ints
+        err = segy_get_field_int( header, field, &f );
         if( err != 0 ) return err;
-        if (ds->lsb) f = bswap_header_word(f, formatsize( fd.datatype ));
+        uint8_t datatype = tr_field_type[field];
```
ajaust: `const uint8_t`?

lib/src/segy.c
```diff
@@ -1006,6 +1057,12 @@ int segy_set_field_int( char* header, const int field, const int val ) {
             fd.value.i8 = (int8_t)val;
             break;
 
+        case SEGY_UNSIGNED_INTEGER_4_BYTE:
+            if ( val < 0 )
```
ajaust, line 1061: Should we also check for `val > UINT32_MAX` here similar to   the other cases like `u16` or `u8` or is this not necessary?

lib/src/segy.c
```diff
@@ -1097,7 +1154,8 @@ int segy_field_forall( segy_datasource* ds,
         // note: for the moment function still works only on ints
         err = segy_get_field_int( header, field, &f );
         if( err != 0 ) return err;
-        uint8_t datatype = tr_field_type[field];
+        uint8_t entry_type = traceheader_default_map[zfield].entry_type;
```
ajaust: For my better understanding. `zfield` is defined as `field-1`. Would this not be an offset (zero-based index) in our new naming scheme or does the shift by 1 has a different meaning here?

lib/src/segy.c
```diff
@@ -1117,7 +1175,7 @@ static int bswap_bin( char* xs, int lsb ) {
     const int bytes4_len = sizeof(bytes4) / sizeof(int);
 
     for( int i = 0; i < bytes4_len; ++i ) {
-        const int offset = bytes4[ i ] - (HEADER_SIZE+1);
+        const int offset = bytes4[i] - ( SEGY_TEXT_HEADER_SIZE + 1 );
```
ajaust: The spaces around `i` disappeared, i.e., `[ i ]` became `[i]`.

lib/src/segy.c
```diff
@@ -1154,7 +1212,7 @@ static int bswap_bin( char* xs, int lsb ) {
     const int bytes2_len = sizeof( bytes2 ) / sizeof( int );
 
     for( int i = 0; i < bytes2_len; ++i ) {
-        const int offset = bytes2[ i ] - (HEADER_SIZE+1);
+        const int offset = bytes2[i] - ( SEGY_TEXT_HEADER_SIZE + 1 );
```
ajaust: The spaces around `i` disappeared, i.e., `[ i ]` became `[i]`.

python/segyio/segyio.cpp
ajaust, line 1727: I just realised after writing that this error message is the error message that we already had (same for errors below). I will just leave the comment here because I am interested to hear what you think. 

It says that the buffer is too small, but couldn't the buffer be too large as well? 

From my understand we only accept buffers of 400-byte (binary header) and 240-byte (some trace header) size. I would prefer an error message to point to this, e.g., `"Unexpected buffer size"`. It could even be more specific and mention `"Unexpected buffer size. Only accept buffers of size 400 byte (binary header) and 240 byte (trace header)"` or include the buffer size `"Unexpected buffer size (%d byte). Only accept buffers of size 400 byte (binary header) and 240 byte (trace headers)"`

, but I can see that this may be too wordy.

If wordyness is not an issue, we could even go

lib/src/segy.c
```diff
@@ -222,159 +222,158 @@ static void bswap16_mem( char* dst, const char* src ) {
...
-    [SEGY_TR_SOURCE_MEASURE_EXP     ] = SEGY_SIGNED_SHORT_2_BYTE,
-    [SEGY_TR_SOURCE_MEASURE_UNIT    ] = SEGY_SIGNED_SHORT_2_BYTE,
-    [SEGY_TR_UNASSIGNED1            ] = SEGY_SIGNED_INTEGER_4_BYTE,
-    [SEGY_TR_UNASSIGNED2            ] = SEGY_SIGNED_INTEGER_4_BYTE,
+ * Default traceheader offset-to-entry-definition map. May be overwritten by
+ * mapping in the file. Possible mapping offsets are in range [0-240), with
+ * first defined offset being positioned at 0. All offsets not explicitly set
+ * are implicitly mapped to {SEGY_ENTRY_TYPE_UNDEFINED, false}.
+ */
+static const segy_entry_definition traceheader_default_map[SEGY_TRACE_HEADER_SIZE] = {
```
ajaust, line 230: I did not check if anything was changed besides subtracting `-1`. Should I go thouroughly through these new offsets once and check them?

lib/src/segy.c
```diff
@@ -1110,57 +1102,73 @@ int segy_get_tracefield( const char* header,
                          int field,
                          segy_field_data* fd ) {
 
-    return get_field( header, mapping, field, fd);
+    int offset = field - 1;
+    int mapsize = SEGY_TRACE_HEADER_SIZE;
```
ajaust: 👍 I like the extra declaration of offset and mapsize. This makes it easier to understand the meaning of `field-1` and the `*HEADER_SIZE` in the context of the function.

I think we could add more `const` here and in the equivalent declarations below. We don't have to if this breaks the style though.

lib/src/segy.c
```diff
@@ -1110,57 +1102,73 @@ int segy_get_tracefield( const char* header,
...
+    int mapsize = SEGY_BINARY_HEADER_SIZE;
+    return get_field_int( header, binheader_map, mapsize, offset, f );
 }
 
 int segy_set_binfield_int( char* header,
                            int field,
                            int val ) {
 
-    return set_field_int( header, binheader_map, field, val );
+    int offset = field - SEGY_TEXT_HEADER_SIZE -1;
```
ajaust:
```suggestion
    int offset = field - SEGY_TEXT_HEADER_SIZE - 1;
```

lib/src/segy.c
```diff
@@ -817,11 +816,59 @@ int segy_close( segy_datasource* ds ) {
     return SEGY_OK;
 }
 
+
+SEGY_FORMAT segy_entry_type_to_format( SEGY_ENTRY_TYPE entry_type ) {
```
yngve793: The current implementation of the switch statement hardcodes the unique mapping from SEGY_ENTRY_TYPE to SEGY_FORMAT. This approach deviates from our established pattern of utilizing tables or enums for static data storage. Given that the new mapping structure we are introducing relies on tables, it feels inconsistent to use a switch statement for what is essentially a static mapping. I recommend considering a more uniform approach that aligns with our existing practices, perhaps by defining this mapping in a table or enum format to enhance maintainability and clarity.

lib/src/segy.c
```diff
@@ -1629,9 +1629,9 @@ int segy_sorting( segy_datasource* ds,
     int len = sizeof( fields ) / sizeof( int );
     for( int i = 0; i < len; ++i ) {
         const int f = fields[ i ];
-        if( f < 0 )
+        if( f < 1 )
```
yngve793, line 1707: Is this an old issue? 
If not it should probably be part of another commit. 
Are there no checks for edge cases?

lib/test/segy.cpp
```diff
@@ -2066,41 +2067,66 @@ TEST_CASE("segy_get_tracefield reads values correctly",  "[c.segy]" ) {
...
 
-    SECTION("test value outside of int16 range for segy_set_tracefield_int") {
+    SECTION( "test value outside of int16 range for segy_set_tracefield_int" ) {
         int32_t value = 0xFFFF + 1;
         Err err = segy_set_tracefield_int( header, SEGY_TR_TRACE_ID, value );
         CHECK( err == Err::value() );
     }
 
-     SECTION("test mismatched type") {
+    SECTION( "test mismatched type for segy_set_tracefield" ) {
```
ajaust: Why doe test the out of range value and mismatched type only for trace fields and not for binary header fields? The tests above tend to have a test for the trace-header operation and one test for the binary-header operation

lib/test/segy.cpp
```diff
@@ -2109,4 +2135,28 @@ TEST_CASE("segy_set_tracefield write invalid value",  "[c.segy]" ) {
...
+
+        err = segy_get_binfield(
+            header, 0, &fd
+        );
+        CHECK( err == Err::field() );
+
+        err = segy_get_binfield(
+            header, SEGY_BINARY_HEADER_SIZE + SEGY_TEXT_HEADER_SIZE + 1, &fd
+        );
+        CHECK( err == Err::field() );
```
ajaust: The tests above tend to have two sections, one for the test for the trace-header operation and one for the binary-header operation. Would it make sense to split this part into two sections as well?

ajaust (commented):
Great job. The changes look good to me. 👍 

I don't have anything serious. I think half of my comments are about `const`ness (which is negotiable) and the rest is for clarifications/better understanding.

applications/segyinfo.c
```diff
@@ -8,10 +8,10 @@
 
 static void printSegyTraceInfo( const char* buf ) {
     int cdp, tsf, xl, il;
-    segy_get_field_int( buf, SEGY_TR_ENSEMBLE, &cdp );
-    segy_get_field_int( buf, SEGY_TR_SEQ_FILE, &tsf );
-    segy_get_field_int( buf, SEGY_TR_CROSSLINE, &xl );
-    segy_get_field_int( buf, SEGY_TR_INLINE, &il );
+    segy_get_tracefield_int( buf, SEGY_TR_ENSEMBLE, &cdp );
```
yngve793, line 11: This issue has be discussed before. The get any field as `int32_t` approach kind of worked in the old code with a couple of exceptions that was handled in each binding via `if` statements. Given the new float fields this has become even more cumbersome. 

Im my opinion there are two ways to handle this in a clean way. 

1. Every field request return a `segy_field_data` struct. In this case the programmer using the code  is responsible for the casting. Advantage is clean and simple c code. Disadvantage more work and responsibility on the programmer using the code. 

2. Every field request uses a data type specific function call. I.e. one get/set function per data type. Advantage idiot proof for the programmer using the code. Disadvantage more functions in the interface. Apparently this an important property. 

Currently we are using a third alternative. A general function for whatever fits into int32_t and spesial cases for the rest. This approach distribute the responsibility for casting, checking data type and using the correct function over both the `c core` library and the programming using the library. Advantage it is very close to the original solution. Disadvantage it fails to take advantage of the new and more general structure introduced in the `c core`.

python/segyio/segyio.cpp
```diff
@@ -1757,13 +1767,18 @@ PyObject* putfield( PyObject*, PyObject *args ) {
...
 
-    if( buffer.len() != SEGY_BINARY_HEADER_SIZE &&
-        buffer.len() != SEGY_TRACE_HEADER_SIZE )
-        return BufferError( "buffer too small" );
-
     int field = (int)PyLong_AsLong(field_arg);
-
     segy_field_data fd;
+
+    switch( buffer.len() ) {
```
yngve793, line 1779: Maybe a discussion for another time. 
It would be better to have a buffer type property that is passes along with the mapping. In this way we would also know that if it was standard, extended or private buffer.

lib/src/segy.c
yngve793: The mapping is defined for a specific header. Thus, the size of the buffer is known and defined for the specific mapping. I find it strange to separate these pieces of information and then add them together as needed. We can structure these data better by defining a struct containing both pieces of information. See https://github.com/yngve793/segyio/commit/a17204fed94ca7bedaae3cbe794e52aabd32a2e2 for an example.

yngve793 (commented):
Nice and structured commits that is pulling us in the right direction. :-)

My comments are generally about structure and how to organise data. No surprise there. :-) Now that we have real commits to start the discussion from, it should be easier to avoid talking past each other. Clearly not everything can be changed at once and some steps are intermediate. Let us discuss a couple of these comments at the next meting.

lib/src/segy.c
achaikou: Generally :+1: , seems to be doable.

Reasoning was to give clarity, which types are actually the same underneath. But consistency approach also makes sense :+1:

lib/src/segy.c
achaikou: This is an old issue. Noticed because analyzer began complaining during some of the intermediate commits.
Valid range of fields is [1, 240], that didn't change.

Only after your comment saw that true, no tests cover that, so added some :+1:

lib/src/segy.c
achaikou: Analyzer complained. Something about "your val is int32, it can not be > than uint32_MAX by definition".

python/segyio/segyio.cpp
achaikou: I hope we would some day introduce a parameter "binary" vs "trace" instead of basing our checks on "field" or "buffer length".

In theory buffer should be allowed to be bigger than 240 or 400.

applications/segyinfo.c
achaikou: I believe we go with alternative 1 as intentional approach :+1: 
Int functions are supposed to legacy code, in my opinion, some note about that was added in the commit itself in header file.

We might leave those for simplicity for current users but kind of "deprecate" them. We ourselves would try to get rid of them gradually.

-- I added explicit comment to each of those functions now. Not sure if we should go with whatever per-compiler deprecation mechanisms exists in C99, especially with our uncertainty regarding those.

python/segyio/segyio.cpp
achaikou: I do not see yet why we would need this property, so yes, up for discussion :cat:

lib/src/segy.c
achaikou: To discuss further.
- mapsize is actually both mapsize and header size... (I just chose mapsize because word size was ambiguous...)
- I think it is a more C-like approach to send array + size instead of combining them into a struct. I think pure arguments are also preferred over structs in segyio.

Generally what you suggest looks like a valid solution, but providing arguments separately is more widely used in C than a struct. I think we need a stronger reason to introduce a struct than this.
So for now not implemented.

lib/src/segy.c
achaikou: I did go through once or twice, so your choice :slightly_smiling_face:

lib/src/segy.c
achaikou: That muse be it! I was wondering what this `zfield` name might be :smile: 

(I do not bother with `segy_field_forall` method for the moment too much.
It still works only on ints and would require certain changes along the way to be updated for different types.)

python/segyio/segyio.cpp
achaikou: Old message, not dealing with it in this commit.

lib/src/segy.c
achaikou: To me consts are just eating already limited line length and are usually more annoying then useful, but can't argue against them in this particular case, so added :smiley: 

~(and no, extra declarations were not for clarity, but due to line length :sweat_smile: )~

lib/test/segy.cpp
achaikou: Before we had one function to deal with both binary and trace fields.
So now as we separated them I felt it is a good idea to check _once_ both bin and trace fields get/set functions.

Once it is checked once, I think we are reasonably safe, given that we know that internally those functions are the same. So I don't feel there is a need to copy tests just for the sake of copying.

If you feel differently - shout. Copying those two lines should be easy. :slightly_smiling_face:

achaikou (commented):
Thanks for the review guys! :smiley:

ajaust (approved):
Thanks for the replies and changes. I have no further comments.

yngve793 (approved):

PR #639, opened 2025-09-12
Start using custom mapping internally

achaikou:
No real custom mapping files or processing, just moving away from default _int functions to proper fd ones in internal library calls.

### DIFF
--- a/lib/include/segyio/segy.h
+++ b/lib/include/segyio/segy.h
@@ -250,6 +250,17 @@ typedef enum {
 } SEGY_ERROR;
 
 
+typedef struct {
+    SEGY_ENTRY_TYPE entry_type;
+    bool requires_nonzero_value; //use only if field value is not zero
+} segy_entry_definition;
+
+typedef struct {
+    char name[8]; //not NULL-terminated
+    uint8_t name_to_offset[SEGY_TRACE_HEADER_SIZE];
+    segy_entry_definition offset_to_entry_definion[SEGY_TRACE_HEADER_SIZE];
+} segy_header_mapping;
+
 /*
  * Represents an abstract data source which supports common file operations,
  * described by function pointers and `writable` attribute.
@@ -333,6 +344,11 @@ struct segy_datasource {
      * certain operations.
      */
     bool memory_speedup;
+
+    /* Standard traceheader mapping. */
+    segy_header_mapping traceheader_mapping_standard;
+    /* Traceheader extension 1 mapping. */
+    //segy_header_mapping traceheader_mapping_extension1;
 };
 
 typedef struct segy_datasource segy_datasource;
@@ -356,11 +372,6 @@ typedef struct {
     segy_field_value value;
     uint8_t datatype;
 } segy_field_data;
-
-typedef struct {
-    SEGY_ENTRY_TYPE entry_type;
-    bool requires_nonzero_value; //use only if field value is not zero
-} segy_entry_definition;
 
 
 /*
@@ -451,6 +462,9 @@ int segy_entry_type_to_datatype( uint8_t entry_type );
 const segy_entry_definition* segy_binheader_map( void );
 /* Default trace header layout map. Indicies (offsets) are 0-based. */
 const segy_entry_definition* segy_traceheader_default_map( void );
+/* Default trace header name to 1-based offset map. ndicies (names) correspond
+ * to SEGY_FIELD enum. */
+const uint8_t* segy_traceheader_default_name_map( void );
 
 /* Reads one trace field data from given 0-based header. 0-based
  * offset-to-entry-definition mapping should correspond to provided header.
--- a/lib/src/segy.c
+++ b/lib/src/segy.c
@@ -2904,34 +2904,95 @@ int segy_binheader_size( void ) {
     return SEGY_BINARY_HEADER_SIZE;
 }
 
-static int scaled_cdp( segy_datasource* ds,
-                       int traceno,
-                       float* cdpx,
-                       float* cdpy,
-                       long trace0,
-                       int trace_bsize ) {
-    int x, y;
+static int scaled_standard_header_cdp(
+    const char* header,
+    const segy_entry_definition* map,
+    int cdp_offset,
+    int scalar_offset,
+    float* cdp
+) {
+
+    segy_field_data fd = {0};
+    int err = segy_get_tracefield( header, map, cdp_offset, &fd );
+    if( err != SEGY_OK ) return err;
+
+    float raw_cdp;
+    const uint8_t entry_type = map[cdp_offset - 1].entry_type;
+
+    switch( entry_type ) {
+        case SEGY_ENTRY_TYPE_IBMFP:
+        case SEGY_ENTRY_TYPE_IEEE32:
+            *cdp = fd.value.f32;
+            return SEGY_OK;
+        case SEGY_ENTRY_TYPE_COOR4:
+        // breaks spec on purpose: scalar is applied anyway
+        case SEGY_ENTRY_TYPE_INT4:
+            raw_cdp = (float)fd.value.i32;
+            break;
+        // breaks spec on purpose: scalar is applied anyway
+        case SEGY_ENTRY_TYPE_UINT4:
+            raw_cdp = (float)fd.value.u32;
+            break;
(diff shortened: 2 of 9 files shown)


yngve793 (approved):
All tests passes. Code looks fine and I do not see any issue.

lib/include/segyio/segy.h
```diff
@@ -250,6 +250,17 @@ typedef enum {
 } SEGY_ERROR;
 
 
+typedef struct {
+    SEGY_ENTRY_TYPE entry_type;
+    bool requires_nonzero_value; //use only if field value is not zero
```
ajaust, line 255: You use different word to describe the meaning of this variable than in the variable itself. Is this an indicator that we should consider using a different name? We could use a name based on the comment, e.g. something like `use_if_field_nonzero`.

How things could look with current name
```c
if (entry.requires_nonzero_value) {
  //Do things
}

if (value != 0 && entry.requires_nonzero_value) {
  //Do things
}
```

vs. name based more on the comment

```c
if (entry.use_if_field_nonzero) {
  //Do things
}


if (value != 0 && entry.use_if_field_nonzero) {
  //Do things
}
```

lib/include/segyio/segy.h
```diff
@@ -250,6 +250,17 @@ typedef enum {
...
+} segy_entry_definition;
+
+typedef struct {
+    char name[8]; //not NULL-terminated
```
ajaust, line 259: Would there be any good reason to store the name as null-terminated string? For example, do we ever expose the name outside of the internal methods to users of the C interface? If we provide this to users anywhere, do we provide it as null-terminated string?

lib/src/segy.c
```diff
@@ -376,14 +375,119 @@ static const segy_entry_definition binheader_map[SEGY_BINARY_HEADER_SIZE] = {
     [ -3201 + SEGY_BIN_NR_TRAILER_RECORDS        ] = { SEGY_ENTRY_TYPE_INT4,   false },
 };
 
+/*
+ * Default traceheader name-to-offset map. Both names and offsets are 1-based. May be
+ * overwritten by mapping in the file. Possible mapping names are in range
+ * [0-240), with first defined name being positioned at 1. All names not
+ * explicitly set are implicitly mapped to offset 0.
+ */
+static const uint8_t traceheader_default_name_map[SEGY_TRACE_HEADER_SIZE] = {
```
ajaust, line 384: We have quite some keywords in names (`default`, `standard`, `map`, `mapping` etc.). I am just wondering if we prefer a certain order and whether the order has special meaning. For example, we have the variable called `segy_traceheader_default_map` and also `traceheader_mapping_standard`. 

- Would `segy_traceheader_map_default` mean something different than `segy_traceheader_default_map`?
- Is there a difference between `map` and `mapping`?

If I understand/remember correctly, some keywords come from the standard (`mapping`, `standard`, `extension1`) and others are our internally used terms (`default`, `map`). Is this roughly correct?

python/segyio/segyio.cpp
```diff
@@ -568,10 +568,56 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
 }
 
-PyObject* segyopen( segyfd* self ) {
+bool overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xline ) {
+    segy_header_mapping& mapping = ds->traceheader_mapping_standard;
+
+    if( py_iline != Py_None ) {
+        long iline = PyLong_AsUnsignedLong( py_iline );
+        if( PyErr_Occurred() || iline < 1 || iline > SEGY_TRACE_HEADER_SIZE ) {
+            ValueError( "Custom iline value out of range" );
```
ajaust: Will users understand that the offset value is out of range and not the value stored at this offset? We could mention `offset` explicitly in the message: 

```suggestion
            ValueError( "Custom iline offset value out of range" );
```

If we decide to change the message here, then we should also change it for the crossline.

python/segyio/segyio.cpp
```diff
@@ -568,10 +568,56 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
+        long xline = PyLong_AsUnsignedLong( py_xline );
+        if( PyErr_Occurred() || xline < 1 || xline > SEGY_TRACE_HEADER_SIZE ) {
+            ValueError( "Custom xline value out of range" );
+            return false;
+        }
+        uint8_t xl = static_cast<uint8_t>( xline );
+        mapping.name_to_offset[SEGY_TR_CROSSLINE] = xl;
+        mapping.offset_to_entry_definion[xl - 1].entry_type = SEGY_ENTRY_TYPE_INT4;
+    }
+    return true;
```
ajaust: The way I understand it, we check if we need to overwrite the mapping. If so, the field offsets are updated in the internally used mapping and we return `true`. However, if both `py_iline` and `py_xline` are `Py_None`, nothing is overwritten, but the methods is still returning `true`. I expected that we would check at whether at least of of the field offsets is overwritten to actually return true.

ajaust (commented):
Looks good overall. I left some minor comments for discussion of names and so I can gain better understanding. 🙂

lib/include/segyio/segy.h
achaikou: I spent a lot of time coming up with this name, so I won't give that up easy :smile: 

"use if field nonzero" is not clear enough for me. This value shows up in layout definition. What does it mean that field is non zero? Does it mean that field byte is non zero? Something else?
"value" helps me here.

But naming variable "use if field value is nonzero" is too long and clumsy. 

But "use if" = "requires". This word gives me the correct context and reads nicer than the alternative and original name. 
Comment is there to help connect field to specification.

lib/include/segyio/segy.h
achaikou: > Would there be any good reason to store the name as null-terminated string?

Don't know.
Decision to go with 8 characters instead of 9 is based purely on my distaste of end NULL and a bit of consistency - our main text headers are allowed to have NULLs inside the text. So we can't treat those as NULL-ending arrays. Thus I wouldn't treat other strings as NULL-ending arrays either.

> do we ever expose the name outside of the internal methods to users of the C interface?

This struct is already public-facing and we do not provide it as NULL-terminated string.

I see though that NULL-terminating strings is a more common C approach. On the other hand, there seems to be some C functions analogues where one must provide size explicitly. Remember seeing a few times that "strncpy is somewhat safer than strcpy". If that is so, why do I need null pointer in the end? I know my size!

Let me know if you are leaning into null-terminated strings instead. :cat:

lib/src/segy.c
achaikou: To me order of words needs to be consistent near similar entities. Otherwise it should read well.
`segy_traceheader_map_default` doesn't read nicely in English as `default` is an adjective, so it usually must go before the `map`.

For `map` and `mapping` I didn't intended the difference consciously (and I see why it might be confusing).
It was probably mostly the context: if I mean table itself, it was named "map". If I mean the concept of something needing some kind of map information to be processed correctly, as in D8, (name to offset map or offset to datatype map or all together), then it was usually "mapping".

You are correct. Words that you described come from specification. "default" and "map" are new ones.

python/segyio/segyio.cpp
achaikou: [And I forgot to add the docs...]

If both values are None and nothing gets overwritten, application state is still good and valid and we don't need to interrupt operations. Note that ValueError doesn't return error, it only sets it. There must be other indicator stating that function has failed. This seems to be the first "helper" function in bindings file, so I chose to go with bool here.

So True here means "all good, application state is fine".

achaikou (commented):
Thanks for the review! :smiley:

lib/include/segyio/segy.h
ajaust: I am fine with either null-terminated or not. I was just not sure how public this information is. As you say, we know the size of this string so there is no need to make it null-terminated. As this struct is part of the public interface and it clearly states the size of the string there is no need to change anything. 😄

ajaust (approved):
Nice! Thanks for the explanations. 🙂

ajaust (approved):
👍 I like the last fixup. With the changes it looks very consistent with the C core library.

PR #640, opened 2025-09-17
docs: adjusted `collect` example

gstvschlz:
The `collect` example was depicting to reshape based on `f.samples` - an array, instead of `len(f.samples)`


ajaust (approved):
Thanks for the fix! 🙂 

I slightly changed the commit message to be in line with the style of the repository and will merge once the tests pass.

PR #642, opened 2025-09-23
D8: Read xml layout

achaikou:
With this PR we can read embedded xml.
For now only works with default mapping. (Different bytes require more changes, different types I have not yet looked at).

- Looks like weird segfaults are defeated (I'll have to see if we could run our python tests with valgrind in CI too, it worked really well).
- I am sure there is still more things to prettify here, but my brain already cannot see them and is counting on yours.
- Special changes for Windows because one job started failing. Probably some update happened...


python/segyio/segyio.cpp
```diff
@@ -439,6 +439,13 @@ struct segyfd {
     int elemsize;
 };
 
+namespace {
+/** Overwrite existing mapping with iline/xline values explicitly provided by
+ * the user. Returns SEGY_OK if requested mapping override was successful, sets
+ * error and returns error code otherwise. */
+int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xline );
+} // namespace
+
```
ajaust, line 504: Why do we only add the forward declaration of `overwrite_mapping` and and the implementation later (line 2090)? Most other methods in this file seem to be immediately implemented at the point of their declaration..

python/segyio/segyio.cpp
```diff
@@ -439,6 +439,13 @@ struct segyfd {
     int elemsize;
 };
 
+namespace {
```
ajaust, line 479: Is the big change here that we move `overwrite_mapping` into an anonymous namespace or is the actual location (i.e. the line number) of the declaration relevant?

lib/src/segy.c
```diff
@@ -3033,3 +3033,83 @@ int segy_rotation_cw( segy_datasource* ds,
     *rotation = (float) radians;
     return SEGY_OK;
 }
+
+static int read_stanza_header( const char* header_start,
```
ajaust: For my understanding: Do I have to read `read_stanza_header` as "Read a stanza from an arbitrary header buffer"?


Reason for the question is that the standard is confusing me. The standard sometimes refers to content of the stanza as "stanza" but other times the content is referred to as "record" and "stanza" refers to this expression `((STANZA))`.

From the standard (about the `EndText` stanza):

> The stanza header shall be on the first line in the record and must be the only non-blank text in the record (i.e., the stanza must be empty).

lib/src/segy.c
```diff
@@ -3033,3 +3033,83 @@ int segy_rotation_cw( segy_datasource* ds,
...
+        *encoding = SEGY_EBCDIC;
+    } else {
+        // block does not start with stanza name, so it must be a continuation
+        // of a previous block
+        *stanza_name_length = 0;
+        return SEGY_OK;
+    }
+    header_start += 2; // skipping ((
+
+    const char* end = memchr( header_start, end_char, header_size );
```
ajaust: Do you know if stanza's of this type are allowed
```text
((NAME: WORD (WORD_IN_PARENTHESES) ANOTHER_WORD))
```
I did not find any part any part of the standard that is explicit about this. Only that parentheses are allowed as part of the `keyword`-`value` pairs of entries within a header.

This does not need to be changes now, but we should be aware of the limitation and document it if we assume that stanza names do not contain any (closing) parentheses.

lib/src/segy.c
ajaust: I am a bit confused about the state of variable when returning from this methods. It greatly depends on where we leave the method. This might be related to my lack of experience with C.

Observation:

1. If the header does not start with `((`, we set the `stanza_name_length` to `0` and return with `SEGY_OK`.
2. If the header starts with `((` but do not find a closing `)`, we set the `encoding` and return `SEGY_NOTFOUND`.
3. If the header starts with `((` and there is a `)` in the buffer, but memory allocation fails, we set the `encoding`, `stanza_name_length`, and `stanza_name` to `NULL` and return `SEGY_MEMORY_ERROR`.
4. If the header starts with `((` and there is a `)` in the buffer and memory allocation succeeds, we set the `encoding`, `stanza_name_length`, and `stanza_name` to some pointer (which one depends on the implementation) and return `SEGY_OK `. The `stanza_name` was reencoded to ASCII (?) if it was encoded in ebcdic initially.

Questions/confusion:
- Why does 1. return `SEGY_OK` instead of `SEGY_NOTFOUND`? We did not find the stanza in the given header.
- Why does 2. not set `*stanza_name_length = 0`, but sets `encoding`?

To me, the cases 1 and 2 are the same in the sense that we did not find a stanza in the current header and thus I first expected to have similar/same state on return.

I see that we do not find/return a stanza for different reasons. In case 1 the header simply does not contain a stanza. In case 2 it looks like there should be a stanza in the header, but it is either over several headers blocks or something else went wrong.

---

Do we want to accept an empty stanza name, i.e., a stanza defined as
```text
(())
```
? This touches upon point 1. from the list above. Would an empty stanza name be legal? At the moment we cannot detect it.

python/segyio/segyio.cpp
```diff
@@ -337,6 +339,20 @@ segy_datasource* create_py_stream_datasource(
 
 } // namespace ds
 
+struct stanza_header {
+    std::string name;
+    int encoding;
+
+    std::string normalize_name() const {
```
ajaust: Maybe call this
```suggestion
    std::string normalized_name() const {
```
to indicate that the method does not normalise, i.e., does not change, the `name` member of the struct.

python/segyio/segyio.cpp
```diff
@@ -337,6 +339,20 @@ segy_datasource* create_py_stream_datasource(
...
+        std::string normalized;
+        for( auto c : this->name ) {
+            if( c != ' ' ) {
+                normalized += std::tolower( static_cast<unsigned char>( c ) );
+            }
+        }
```
ajaust, line 359: We could express this using the STL as
```suggestion
        std::transform(normalized.begin(), normalized.end(), normalized.begin(),
                   [](unsigned char c) { return std::tolower(c); });
```

I don't think we need any special handling of characters like `' '` since `std::tolower` should only affect capital (ASCII) characters.

Not sure if you like it, but people told me too much to use the STL. 😅

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
     return SEGY_OK;
 }
 
+int parse_extended_text_headers( segyfd* self, const char* binheader ) {
+    segy_datasource* ds = self->ds;
+
+    segy_field_data fd;
+    int extra_headers = 0;
+    segy_get_binfield( binheader, SEGY_BIN_EXT_HEADERS, &fd );
+    extra_headers = fd.value.i16;
```
```suggestion
    segy_get_binfield( binheader, SEGY_BIN_EXT_HEADERS, &fd );
    const int extra_headers = fd.value.i16;
```
This would save one line. 😄

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
...
+        int encoding = ds->encoding;
+
+        int err = segy_stanza_header(
+            ds, i, &c_stanza_name, &stanza_name_length, &encoding
+        );
+        if( err != SEGY_OK ) return err;
+
+        if( stanza_name_length == 0 && i > 0 ) {
+            encoding = self->stanzas[i - 1].encoding;
+        }
```
ajaust: Just an alternative way to write it to avoid indices as much as possible 

```suggestion
        if( stanza_name_length == 0 && not self->stanzas.empty() ) {
            encoding = self->stanzas.last().encoding;
        }
```

I am not sure if I like it more (yet) because the notion of "previous" (`i-1`) is less obvious.

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
...
+
+        stanza_header stanza = {
+            std::string( c_stanza_name, stanza_name_length ),
+            encoding
+        };
+        self->stanzas.push_back( stanza );
+
+        if( c_stanza_name ) free( c_stanza_name );
+        if( stanza.normalize_name() == "seg:endtext" ) break;
```
ajaust: You could write it slightly differently which might avoid one copy of the string (the copy is probably optimised away by the compiler already anyway). I had to mention it. because I am proud that I remember `emplace_back`.


```suggestion
        self->stanzas.emplace_back(
            std::string( c_stanza_name, stanza_name_length ),
            encoding
        );

        if( c_stanza_name ) free( c_stanza_name );
        if( stanza.back().normalize_name() == "seg:endtext" ) break
```

python/segyio/segyio.cpp
ajaust: Do you think we should safe guard against values that are not defined, e.g., if the value is `-2` instead of `-1` or do we always assume that the file is healthy anyway (garbage in, garbage out) and thus don't need such fine-grained checks?

lib/src/segy.c
```diff
@@ -988,54 +988,57 @@ static int get_field( const char* header,
...
 
-    if ( offset < 0 || offset >= mapsize ) {
+    if( offset < 0 || offset >= mapsize ) {
         return SEGY_INVALID_FIELD;
     }
 
-    fd->datatype = entry_type_to_datatype_map[mapping[offset].entry_type];
+    fd->entry_type = mapping[offset].entry_type;
+    uint8_t datatype = entry_type_to_datatype_map[fd->entry_type];
+    int vsize = formatsize( datatype );
```
ajaust, line 997: Is `vsize` short for "value size"?

lib/src/segy.c
```diff
@@ -1770,7 +1777,7 @@ int segy_sample_interval( segy_datasource* ds, float fallback, float* dt ) {
     segy_get_binfield( bin_header, SEGY_BIN_INTERVAL, &fd );
     bindt = fd.value.i16;
 
-    segy_entry_definition* standard_map =
+    const segy_entry_definition* standard_map =
```
ajaust, line 1780: 😍

python/segyio/segyio.cpp
```diff
@@ -1862,7 +1863,7 @@ PyObject* getfield( PyObject*, PyObject *args ) {
             return PyFloat_FromDouble( fd.value.f64 );
 
         default:
-            return KeyError( "Unhandled datatype %d for field %d", fd.datatype, field );
+            return KeyError( "Unhandled datatype %d for field %d", fd.entry_type, field );
```
ajaust: Should this either be 
```suggestion
            return KeyError( "Unhandled datatype %d for field %d", datatype, field );	  
```
or be rephrased as
```suggestion
            return KeyError( "Unhandled entry_type %d for field %d", fd.entry_type, field );  
```
?

python/segyio/segyio.cpp
```diff
@@ -2150,7 +2150,7 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
         }
         uint8_t il = static_cast<uint8_t>( iline );
         mapping.name_to_offset[SEGY_TR_INLINE] = il;
-        mapping.offset_to_entry_definion[il - 1].entry_type = SEGY_ENTRY_TYPE_INT4;
+        mapping.offset_to_entry_definion[il - 1] = { SEGY_ENTRY_TYPE_INT4, false, nullptr };
```
ajaust: I tried to figure out if `NULL` (from C) and `nullptr` (from C++) are always save to mix. At least, [they are not always the same.](https://stackoverflow.com/questions/36484473/is-c-null-equal-to-c11-nullptr)

As we overwrite the mapping created in the C interface, we have to keep an eye on this what happens, e.g., if we use different compilers for C and C++. As this data structure is passed to C code, using `NULL` should be safest if we are uncertain.

Do you know more?

python/segyio/segyio.cpp
```diff
@@ -467,6 +468,9 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
  * stanza headers. Updates stanzas field. */
 int parse_extended_text_headers( segyfd* self, const char* binheader );
 
+/** Frees segy_header_mapping names allocated on heap. */
+int free_header_mappings_names( segy_header_mapping* mappings, size_t mappings_length );
```
ajaust, line 485: Why do we use `_length` over `_size`? I have the impression that we tend to use `size` for the size/length of data structures. `length` is used for quantities/entries that come from the standard like the inline length, crossline length, trace length, or similar.

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+        }
+    }
+    return SEGY_OK;
+}
+
+struct NameMapEntry {
+    std::string spec_entry_name;
+    int segyio_entry_name;
+};
+struct TypeMapEntry {
```
ajaust, line 2351: Missing empty line between the structs.

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+    std::string spec_entry_type;
+    SEGY_ENTRY_TYPE segyio_entry_type;
+};
+
+/** D8 names to SEGY names. Contains all the known names, not only the ones used
+ * internally in the library. scale6 type entries are not fully supported as we
+ * have separate names for mantissa and exponent parts of the type. */
+static const NameMapEntry standard_name_map[] = {
```
ajaust, line 2359: I did not check them individually. If I should do this, please let me know.

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+    { "elev4",    SEGY_ENTRY_TYPE_ELEV4       },
+    { "time2",    SEGY_ENTRY_TYPE_TIME2       },
+    { "spnum4",   SEGY_ENTRY_TYPE_SPNUM4      },
+    { "scale6",   SEGY_ENTRY_TYPE_SCALE6_MANT },
+};
+
+int set_mapping_name_to_offset(
+    segy_header_mapping* mapping,
+    std::string spec_entry_name,
+    int byte
```
ajaust, line 2476: I am not completely sure where our "name boundary" is. From what point in our code do we call the offset `offset` and when `byte` (as in byte offset). I was surprised to read `byte` instead of `offset` in this commit at a few places.

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+            return SEGY_INVALID_ARGS;
+        }
+        segy_header_mapping mapping = {};
+
+        Py_ssize_t header_size;
+        const char* header_name = PyUnicode_AsUTF8AndSize( py_header_name, &header_size );
+        if( !header_name || header_size > 8 ) {
+            return SEGY_INVALID_ARGS;
+        }
+        strncpy( mapping.name, header_name, 8 );
```
ajaust, line 2598: - Is it (theoretically) possible that `header_size` is < 8? 
  - Would this a problem? 
  - Should we check for `header_size != 8`?

lib/src/segy.c
```diff
@@ -3033,3 +3033,83 @@ int segy_rotation_cw( segy_datasource* ds,
...
+                        int* stanza_encoding ) {
+    int offset =
+        SEGY_TEXT_HEADER_SIZE + SEGY_BINARY_HEADER_SIZE + SEGY_TEXT_HEADER_SIZE * i;
+    char header[SEGY_TEXT_HEADER_SIZE];
+
+    if( !ds->minimize_requests_number ) {
+        int err = ds->seek( ds, offset, 0 );
+        if( err != 0 ) return SEGY_DS_SEEK_ERROR;
+
+        const int initial_read_size = 128;
```
yngve793: Should this be a defined constant with upper case letters?

python/segyio/segyio.cpp
```diff
@@ -2110,6 +2087,36 @@ PyMethodDef SegyMethods[] = {
     { NULL }
 };
 
+namespace {
+
+int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xline ) {
+    segy_header_mapping& mapping = ds->traceheader_mapping_standard;
```
achaikou: TODO somewhere and somehow:
mapping must be overwritten in both c++ and c copies. Here we overwrite only the C version, not values from `self->traceheader_mappings`.
Duplication happens already in `set_traceheader_mappings`, so before this call.

Not sure yet what would be the best way to deal with this is, given that in no-xml case default ds is copied to traceheader_mappings and in xml-case custom traceheader_mappings is copied to ds.

python/segyio/segyio.cpp
yngve793: Declaration of "extra_headers" can be done in the same line. 
`int extra_headers = fd.value.i16;`

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
...
+    int extra_headers = 0;
+    segy_get_binfield( binheader, SEGY_BIN_EXT_HEADERS, &fd );
+    extra_headers = fd.value.i16;
+
+    int i = 0;
+    while( extra_headers != i ) {
+        char* c_stanza_name = NULL;
+        size_t stanza_name_length;
+        int encoding = ds->encoding;
+
```
yngve793, line 2304: Why dont we also store the data and the data length here?

python/test/segyio_c.py
```diff
@@ -648,3 +648,29 @@ def test_memory_buffer_memory_management_on_uncaught_error():
...
+    assert f.stanza_names() == ['', '', '', '']
+    f.close()
+
+    filename = str(testdata / 'stanzas-known-count.sgy')
+    f = segyfile(filename, "r", 0)
+    f.mmap()
+    f.segyopen()
+    big = 'SEGYIO:TEST ASCII  DATA WITH CONTENTTYPE AND BYTES: application/vnd.openxmlformats-officedocument.wordprocessingml.document.glossary+xml:666'
+    assert f.stanza_names() == [
+        big, 'SEGYIO:Test EBCDIC data', 'SEGYIO: test ASCII data']
```
yngve793: Suggestion: Drop the definition of "big" and put one string per line.

lib/include/segyio/segy.h
```diff
@@ -869,6 +869,18 @@ int segy_crossline_stride( int sorting,
...
                            int* stride );
 
+
+/*
+ * Retrieves stanza name and encoding from stanza header i (0-based starting
+ * from first extended text header). Memory is allocated for stanza name and
+ * must be freed by the caller.
+ */
+int segy_stanza_header( segy_datasource* ds,
+                        int i,
```
yngve793: User a name for the stanza heder number that is more aligned with the rest of the parameters. For instance "stanza_nr". A single "i" is usually reserved for simple iteration over some elements inside a loop.

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
     return SEGY_OK;
 }
 
+int parse_extended_text_headers( segyfd* self, const char* binheader ) {
+    segy_datasource* ds = self->ds;
+
+    segy_field_data fd;
+    int extra_headers = 0;
```
yngve793: It would be nice to have a "_nr" as prefix or post fix to indicate that this is a number.

python/segyio/segyio.cpp
ajaust: I don't have the full overview yet, but it indeed sounds potentially troublesome that we do not have a unified flow for how the mapping is set.

python/segyio/segyio.cpp
```diff
@@ -2116,6 +2142,42 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
...
+    segy_get_binfield( binheader, SEGY_BIN_EXT_HEADERS, &fd );
+    extra_headers = fd.value.i16;
+
+    int i = 0;
+    while( extra_headers != i ) {
```
ajaust: What is the reason that we use a `while` loop over a `for` loop here?

python/segyio/segyio.cpp
```diff
@@ -2161,7 +2161,7 @@ int overwrite_mapping( segy_datasource* ds, PyObject* py_iline, PyObject* py_xli
         }
         uint8_t xl = static_cast<uint8_t>( xline );
         mapping.name_to_offset[SEGY_TR_CROSSLINE] = xl;
-        mapping.offset_to_entry_definion[xl - 1].entry_type = SEGY_ENTRY_TYPE_INT4;
+        mapping.offset_to_entry_definion[xl - 1] = { SEGY_ENTRY_TYPE_INT4, false, nullptr };
```
ajaust: If we choose `NULL` over `nulltpr`, then we should update this line accordingly.

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+int free_header_mappings_names(
+    segy_header_mapping* mappings,
+    size_t mappings_length
+) {
+    if( !mappings ) return SEGY_OK;
+
+    for( size_t i = 0; i < mappings_length; ++i ) {
+        segy_header_mapping* mapping = &mappings[i];
+        for( int j = 0; j < SEGY_TRACE_HEADER_SIZE; ++j ) {
+            char*& name = mapping->offset_to_entry_definion[j].name;
```
ajaust: I just realised that there is a typo in our definition of the `segy_header_mapping` struct. I think this should read as:
```suggestion
            char*& name = mapping->offset_to_entry_definition[j].name;
```

python/segyio/utils.py
```diff
@@ -131,3 +132,54 @@ def make_segyfile_descriptor(self, endian, encoding):
             encoding=c_encoding(encoding),
         )
         return fd
+
+
+class TraceHeaderLayoutEntry:
```
ajaust, line 137: Should this be a `dataclass`?

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+        err = set_mapping_offset_to_entry_defintion(
+            mapping, spec_entry_name, byte, spec_entry_type, requires_nonzero_value
+        );
+        if( err != SEGY_OK ) return err;
+    }
+    return SEGY_OK;
+}
+
+int parse_py_traceheader_layout_dict(
+    PyObject* headers,
```
ajaust, line 2581: Is the intended `dict` structure like this

```
{
  "SEG00001": [
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
  ]
  "MYHEADR0": {
      TraceHeaderLayoutEntry(name, offset, type, requires_nonzero_value),
    },
}
```
?

python/segyio/segyio.cpp
```diff
@@ -2202,6 +2206,338 @@ int parse_extended_text_headers( segyfd* self, const char* binheader ) {
...
+    }
+    return SEGY_OK;
+}
+
+extern "C" int segy_parse_layout_xml(
+    const char* xml,
+    size_t xml_size,
+    segy_header_mapping** mappings,
+    size_t* mappings_length
+) {
```
ajaust, line 2613: For my understanding: We have the XML parsing in Python and call it from C. This makes it mandatory to compile segyio with `BUILD_PYTHON=ON` if one intends to use the C interface with user-defined headers. In this case, Python would become an additional dependency.

python/segyio/segyio.cpp
```diff
@@ -2538,6 +2564,102 @@ extern "C" int segy_parse_layout_xml(
...
+
+int extract_xml_from_layout_stanza(
+    segyfd* self,
+    std::vector<char>& xml_stanza_data
+) {
+    segy_datasource* ds = self->ds;
+
+    auto is_layout_stanza = []( const stanza_header& s ) {
+        const std::string layout_stanza_name = "seg:layout";
+        return s.normalize_name().compare( 0, layout_stanza_name.size(), layout_stanza_name ) == 0;
```
ajaust: I like the lambda function as it gives the comparison a good, readable name. I think we might be able to simplify if I understand correctly what it is doing:
```suggestion
        return s.normalize_name() == "seg:layout";
```

python/segyio/segyio.cpp
```diff
@@ -2538,6 +2564,102 @@ extern "C" int segy_parse_layout_xml(
...
+    if( it == self->stanzas.end() ) return SEGY_OK;
+
+    size_t start_index = std::distance( self->stanzas.begin(), it );
+    size_t end_index = self->stanzas.size();
+    for( size_t i = start_index + 1; i < self->stanzas.size(); ++i ) {
+        if( self->stanzas[i].normalize_name() != "" ) {
+            end_index = i;
+            break;
+        }
+    }
```
ajaust: Above you use a lambda function and `find_if`. Couldn't you do the same thing here? Something like (I did not test it):

```suggestion
    auto is_stanza = [](const stanza_header& s){
        return s.normalize_name() != "";
    }
    auto end_it = find_if(
        it, self->stanzas.end(), is_stanza
    );
```

I think it would be nice to do it twice using the same approach.

test-data/README.md
```diff
@@ -84,3 +84,18 @@ Dimensions are mentioned in the order from slowest to fastest changing.
...
+Files are created manually.
+
+| File                           | Purpose                                                       |
+|--------------------------------|---------------------------------------------------------------|
+| mapping-default.sgy            | XML corresponds exactly to the default SEG-Y rev 2.1 mapping. |
+| mapping-invalid-attribute.sgy  | Last entry does not have 'byte' attribute.                    |
+| mapping-invalid-root.sgy       | Root does not define standard header layout.                  |
+| mapping-invalid-value.sgy      | Last standard header entry byte is out of range.              |
+| mapping-multiple-stanzas.sgy   | Other stanzas are present before and after layout stanza.     |
+| mapping-unparsable.sgy         | XML is unparsable.                                            |
```
ajaust, line 100: Should we also add a test where the mapping is (i) valid and (ii) different from the default? Like a file where inline and crossline position are swapped or at a non-standard location?

ajaust (commented):
I don't think I had any major remarks. I have some questions for clarification and some comments w.r.t. to C++ stuff. The C++ remarks are mostly ideas how one could do it differently, but you have to see if you see the benefit in doing it the suggested way.

python/segyio/segyio.cpp
ajaust: I just realized that it probably should by `self->stanzas.back()` instead of `self->stanzas.last()`.

python/segyio/segyio.cpp
achaikou: Location (line number).
Having it down in the file doesn't pollute the reading logic from class to init to segy open.

python/segyio/segyio.cpp
achaikou: Because all the related functions would take many lines of code, but their content is not that interesting.

general structs - `segyfd` - `init` - `segyopen` are the backbone of this class. This is the code you usually scroll up to and check many times during development.

Keeping implementations of layout-related stuff here cluttered that flow too much as those are just implementation details, while structs, `init` and `segyopen` describe the vital flow of file opening.

python/segyio/segyio.cpp
achaikou: Did something as a fixup for the last commit.
Probably could be tested later where both xml and iline-xline override default values to something else.

lib/src/segy.c
achaikou: In this case by `read_stanza_header` I meant `"read ((????))"`

Some quotes:

> Stanza begins on a 3200-byte boundary with a **stanza header**, which is a line containing the name of the defining organization or company and the name of the stanza, separated by a colon.

> The **stanza header** begins with double left parentheses

> Any content following the closing parentheses of the **Stanza Name** is considered to be part of the **Stanza data**.

So for my understanding
- "stanza" is all extended header bytes that correspond to some type of information.
- "stanza header" is `((NAME))`. In many parts of the spec they also call header "stanza name", but I used "name" for part without `((` and `))`
- "stanza data" is actual information user wants to convey, without stanza header.

I renamed "header" variables to make it more clear.

achaikou: I added some internal docs, hopefully it will help with overview a bit.

1. This case means that header is NOT a new stanza, but a continuation of a previous one. Which means there is no stanza header. So this is not an error, which means SEGY_OK is fine. No special handling is required.
2. Means that either provided text buffer is too small or something is wrong. It is an error, user must deal with it. Encoding is set just because it was convenient. Main point here is that error is returned, so one should assume that all retrieved data can be invalid.
3. If memory allocation fails, we do not set anything to NULL, we just return an error. All output data can be garbage.
4. If name is found and memory allocated without problems. Name is reencoded to ASCII the same way we re-encode data in `segy_read_ext_textheader` - externally our library always operates on ASCII.

Empty stanza name is stupid, but not illegal enough for us to fail reading. We would just add an empty name to the list, from there it is on the user to decide what that is.

lib/src/segy.c
achaikou: Good point! I think I didn't find in-build function that would find me `))` and then forgot about it.

It is easy enough to add, so something is done. It is not 100% robust, but lets do more only when someone complains.

python/segyio/segyio.cpp
achaikou: I think I was considering using `std::transform` but decided that simple loop is easier.

Problem: your code does not normalize the string - it does not remove spaces inside.
 

> To aid readability, spaces (" ", EBCDIC 4016 or ASCII 2016) within stanza names shall be allowed but ignored. Thus the stanza name (( SEG: Recording Parameters )) shall refer to the same stanza as ((seg:RECORDINGPARAMETERS)).

python/segyio/segyio.cpp
achaikou: Valid concern...

Because we are not the ones writing the files, it is hard to tell if failing whole open because there is garbage in place of "extended headers" is a good idea. Standard gives no meaning to negative numbers other than -1, but maybe writers will add it?

So I like your definition of "garbage in - garbage out" :smiley:

python/segyio/segyio.cpp
achaikou: I find `while` more natural in this case because `extra_headers` can be -1.

`for (int i = 0, i != -1, ++i)` with `break` inside
seems more confusing to me than
`while( -1 != i )` with `break` inside.

lib/src/segy.c
achaikou: Yes, that was the intention :)

python/segyio/segyio.cpp
achaikou: No, I do not know more, but from your link
> In C++11 and beyond, a pointer that is ==NULL will also ==nullptr and vice versa.

I agree though that using NULL here would not harm :+1:

python/segyio/segyio.cpp
achaikou: Had a long discussion with Copilot about that :smile: 
Came to a conclusion that it wouldn't matter in many cases whether we use size or length.

In this case `mappings_size` seems a bit less readable (maybe I am talking about total byte size?).
length seemed to state more clearly: "this is the length of the list".

python/segyio/segyio.cpp
achaikou: :scream_cat: 
:see_no_evil: 
Would change as a separate commit...

python/segyio/segyio.cpp
achaikou: This part of code deals directly with xml, where it is called byte.
So I think it makes sense to call it `byte` here, but `offset` once we leave xml and don't care anymore if data came from xml or any other place.

python/segyio/segyio.cpp
achaikou: Yes. It is dict of lists :+1:

python/segyio/segyio.cpp
achaikou: Header size < 8 bytes should be a valid thing and I believe we deal with it correctly.

From the `PyUnicode_AsUTF8AndSize` [docs](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize)

> The returned buffer always has an extra null byte appended (not included in size), regardless of whether there are any other null code points.

 
From [strncpy](https://en.cppreference.com/w/cpp/string/byte/strncpy.html)

> Copies at most count characters of the byte string pointed to by src (including the terminating null character) to character array pointed to by dest.
> 
> If count is reached before the entire string src was copied, the resulting character array is not null-terminated.
> 
> If, after copying the terminating null character from src, count is not reached, additional null characters are written to dest until the total of count characters have been written.


So if header name was "Cat", we should end up with
`Cat\0\0\0\0\0`
Will be able to test that only when we get to custom proprietary headers though.

python/segyio/segyio.cpp
achaikou: Hmmm, I didn't think that pure-C-library users would be able to use user-defined headers this way. My knowledge of compilation is not good enough to answer that.

My idea was that if pure-C-library users want to deal with user-defined headers, they will have to implement `segy_parse_layout_xml` on their own. (At least for now, before we figure out the best way to implement this in C library ~which may never happen~).

python/segyio/utils.py
achaikou: Not sure how much it would be of use, given that `__init__` would have to be redefined anyway because of the type conversions.

But generally have nothing against it and would implement if you think it is nicer :)

python/segyio/segyio.cpp
achaikou: I checked them once or twice, so skip if you wish :)

python/segyio/segyio.cpp
achaikou: Look at the [seg-y rev 2.1 clarification](https://library.seg.org/pb-assets/technical-standards/seg-y_rev2-1_clarification_2025-05-01.pdf) 
name could be for example
`SEG:LAYOUT:text/xml`

So we need to do `compare` or similar as we don't have `starts_with` yet in c++11 (and this is probably not a good enough reason for upgrade).

Other option would be to try cutting the ending in normalization, but just `compare` seems simpler than finding correct `:` to cut on.

python/segyio/segyio.cpp
achaikou: Yes, that is nicer! Totally didn't occur to me :+1:

test-data/README.md
achaikou: Not in purpose for this PR as its support requires more code changes.
> For now only works with default mapping. (Different bytes require more changes, different types I have not yet looked at).

achaikou (commented):
Great! Thanks for the review! :cat:

lib/src/segy.c
```diff
@@ -3033,3 +3033,83 @@ int segy_rotation_cw( segy_datasource* ds,
...
+        // block does not start with stanza name, so it must be a continuation
+        // of a previous block
+        *stanza_name_length = 0;
+        return SEGY_OK;
+    }
+    header_start += 2; // skipping ((
+
+    const char* end = memchr( header_start, end_char, header_size );
+    if( end != NULL ) {
+        *stanza_name_length = (size_t)( end - header_start );
```
yngve793: I cant seam to find the check for the second closing bracket.

lib/src/segy.c
```diff
@@ -3033,3 +3033,83 @@ int segy_rotation_cw( segy_datasource* ds,
...
+        }
+    } else {
+        // not enough data to retieve stanza name
+        return SEGY_NOTFOUND;
+    }
+
+    return SEGY_OK;
+}
+
+int segy_stanza_header( segy_datasource* ds,
```
yngve793: I am struggling to understand the thought behind the approach here.
Why do we move to text header number "i" and only focus on the stanza_name? 
The hole buffer is definitely in memory by now and we have the uncertainty of the name length. Why not parse the hole text header and return the relevant information like name, text data, encoding etc?

python/segyio/segyio.cpp
```diff
@@ -337,6 +339,20 @@ segy_datasource* create_py_stream_datasource(
 
 } // namespace ds
 
+struct stanza_header {
+    std::string name;
+    int encoding;
```
yngve793: Why not add text body here?

lib/src/segy.c
```diff
@@ -988,54 +988,57 @@ static int get_field( const char* header,
                       int offset,
                       segy_field_data* fd ) {
 
-    if ( offset < 0 || offset >= mapsize ) {
+    if( offset < 0 || offset >= mapsize ) {
```
yngve793, line 991: Is space change intended?

python/segyio/segyio.cpp
```diff
@@ -600,6 +623,9 @@ PyObject* segyopen( segyfd* self, PyObject* args, PyObject* kwargs ) {
     int err = segy_binheader( ds, binary );
     if( err ) return Error( err );
 
+    err = parse_extended_text_headers( self, binary );
```
yngve793: Why do we pass a buffer of size 400 [SEGY_BINARY_HEADER_SIZE] to the parse_extended_text_headers?

python/segyio/segyio.cpp
```diff
@@ -2538,6 +2564,102 @@ extern "C" int segy_parse_layout_xml(
...
+    size_t start_index = std::distance( self->stanzas.begin(), it );
+    size_t end_index = self->stanzas.size();
+    for( size_t i = start_index + 1; i < self->stanzas.size(); ++i ) {
+        if( self->stanzas[i].normalize_name() != "" ) {
+            end_index = i;
+            break;
+        }
+    }
+
+    size_t headers = end_index - start_index;
```
yngve793: First half of the `extract_xml_from_layout_stanza` is very had to follow and get an intuition for what we are tying to achieve and why. My best guess is that we are calculating the number of extended text headers used to define the xml.  As comments is not an option, could we move this into a function with a descriptive name?

yngve793 (approved):
Out of 12 comment is only seams to register 2 when writing this summary. Let us hope that Github works better behind the behind the frontend. 

All comments are written independent of comments from @ajaust, thus there might be repeated comments or issues that are already resolved in fixups. 

I struggle a bit to follow the over all thought in the c++ and c part of the code. Overall it looks fine, but confidence is at medium. Creating a propper suggestion will imply solving the problem all over again. 

Hopefully some of the comments will be useful.

lib/src/segy.c
achaikou: I think I either forgot it originally or decided that it is unlikely to happen :blush: 
It is now added as a fixup :cat:

lib/src/segy.c
achaikou: No, because this is a local const variable.
As far as I understand, uppercase usually goes for some global constants.

python/segyio/segyio.cpp
achaikou: Original intention was to read just a few bytes from each header, simply to create a header name map, because this is information that is needed.

I am sure that almost in 100% of cases we won't read more than 128B of each 3200B.
And the size of all extended headers in unknown and depends on the file. From PO we know that there is an example where this could be 22MB.
I believe we should keep `open` as fast as we can. Depending on the datasource, difference between reading 22 MB (all text) and 1 MB (just enough text to find the name), might matter.

Extended headers themselves are not usually required for us and segyio has a policy of not caching stuff. (Even the binary header is not cached!!).
As a rule, information is retrieved only when user asks for it.

So from this point of view I don't see why we should store extended headers themselves.

We can discuss that further if you think that we should read all the headers and save their information in memory. :cat:

python/segyio/segyio.cpp
achaikou: Because number of extended text headers is possibly written in the binary buffer, so we would read value from it.

I based this signature on `segy_samples`, `segy_format`, etc, but it doesn't have to be like that.
Maybe we can drop it and just read binary header inside the function.
But what is better in this case probably depends on what we decide to do in another PR...

python/segyio/segyio.cpp
achaikou: Intention is not to read the full header. Keeping it in memory doesn't give any obvious benefits also.

For detailed explanation see other comment :cat:

lib/src/segy.c
achaikou: > The hole buffer is definitely in memory by now

Hopefully only 128 bytes are in memory.

For detailed explanation see other comment.

lib/src/segy.c
achaikou: Yes, my formatter complained :smile:

python/segyio/segyio.cpp
achaikou: Is it better now with suggested change from Alex to use lambda? Or do you think we still need the function? If so, will do!

(And yes, your best guess is correct!)

lib/include/segyio/segy.h
achaikou: In segy_read_ext_textheader parameter was called pos, I didn't like it and decided "i" is more understandable. Apparently not :smile: 

Changed to `headerno` to be aligned with `traceno` used on traceheaders operation.

python/segyio/segyio.cpp
achaikou: Used "count" as this is a standard segyio prefix for such things.

achaikou (commented):
Thanks for the review!
(And of course it was useful! :cat: )

python/segyio/segyio.cpp
yngve793: Only the SEGY_BIN_EXT_HEADERS value is used. Why not pass the value it self over the pointer to the buffer. Tricks like this makes the code unnecessary hard to read.

python/segyio/segyio.cpp
yngve793: If you keep the start and end or start and length of the data it would simplify the `extract_xml_from_layout_stanza` function. In the `extract_xml_from_layout_stanza` function simply count up the exact number of bytes and read with the correct offset without erasing data at a later point.

python/segyio/segyio.cpp
yngve793: I suggest the following:
- Structure the code better. Keep related lines close to each other. 
- Part 1: Find first occurrence of "seg:layout".  
- Part 2: Find last occurrence of "" 
- Part 3: Calculate the correct size of `xml_stanza_data`
- Part 4: Read the exact data block from the header

It is a bit unclear why two iterators are used here as we continuing trough the same vector.

python/segyio/segyio.cpp
achaikou: OK, then I would prefer to have signature 
`parse_extended_text_headers( self );`
(which I changed to now, hopefully it solves your problem with code not being easy to read)

`SEGY_BIN_EXT_HEADERS` is just a field, it is not really number of extended headers.
So if we would not follow the pattern of other existing functions, it makes more sense to read this value inside the method as a step in figuring out extended headers, than to provide it directly from the outside.

python/segyio/segyio.cpp
achaikou: I try to guess what is the rootcause of your comment. I am not sure that I understand you correctly.

Short version:

I was very much focused on compliance with `segy_read_ext_textheader`. After your comment I reconsidered.
Suggested WIP commit is added on top of it and is a subject for discussion.

---
Long version:

Lets say that structure is like that:
- Stanza "ST1" is set to blocks [0; 4)
- Stanza "ST2" is set to block [4, 6)
- End Stanza "EndText" is set to block 6.

In the current design users can only read extended text headers via function `segy_read_ext_textheader` which accepts _index of the header to read_. Whatever user is doing further with this information is not our problem.

Treating extended text headers as stanzas is a new thing and I thought we should not support it thoroughly until PO requests we spend time on it improving user experience. I thought until then users would keep using `segy_read_ext_textheader`.
But, if we had time, we probably could give users a list of information like
```
0 - ["ST1", ASCII]
1 - ["", ASCII]
2 - ["", ASCII]
3 - ["", ASCII]
4 - ["ST2", EBCDIC]
5 - ["", EBCDIC]
6 - ["EndText", ASCII]
```
 so that they know which indices they need to read with `segy_read_ext_textheader` (currently added to `stanza_names` binding function, but not yet exposed to the users).

Internally we need to parse stanzas to 1) find layout stanza 2) find end text stanza.
So intention of this `parse_extended_text_headers` function was to find header names for each extended text header and then internally do the same thing as users: read headers with `segy_read_ext_textheader`.
Function itself in its current form does not know which stanza it belongs to.

So I suspect you suggest to change the outcome of this function from identifying  _indexed by extended header_ stanza-parts to identifying full stanzas? Something like
```
0 - ["ST1", ASCII, headerindex=0, headercount =4]
1 - ["ST2", EBCDIC, headerindex=4, headercount = 2]
2 - ["EndText", ASCII, headerindex = 6, headercount = 1]
```
(I think this is the most readable form. If we want size of data in bytes, without for example name size, we could just add method to stanza_header that would calculate it, as we already have all the information)

Reading direct number of bytes without doing `erase` in `extract_xml_from_layout_stanza` is more complicated because here we have a design clash.
- If we read it directly, we don't have access to `encode` function - which is likely done on purpose because all such things should happen inside C library.
- If we create a new function, per design of other functions we better read full text headers, including the name.

If this is what you mean, then I think you are right and this solution would be better. Maybe my focus on keeping `segy_read_ext_textheader` function was wrong as it makes little sense from the user's point of view.
Though we need to discuss it and probably plan in more time to give users support for stanzas.

python/segyio/segyio.cpp
achaikou: Related to problems explained in another commit, so would need to discuss.

achaikou (commented):
I added a few new commits, not as fixups for at least some clarity. Without proper rebase this PR is a mess anyway. Prettyfing will be thus done later.
Changes are done after conversation with PO and I think somewhere along the lines of your suggestions, Yngve.

lib/src/segy.c
```diff
@@ -3024,49 +3024,73 @@ int segy_rotation_cw( segy_datasource* ds,
...
-                               char** stanza_name,
-                               size_t* stanza_name_length,
-                               int* encoding ) {
+/** Reads stanza header ((stanza_name)) from the text.
+ *
+ * If header is found, stanza name and stanza name length is extracted and set.
+ * Encoding is determined based on the name. stanza_name memory is allocated on
+ * heap. If allocation fails, SEGY_MEMORY_ERROR is returned.
+ *
+ * If starting characters of the header are not found, text is assumed to be
```
ajaust, line 3033:
```suggestion
 * If starting characters of the header are not found, content is assumed to be
```
Since I learned that it does not have to be text. I am fine with "text" as well since it is the "textual" header, but I had to bring it up now. 😅

lib/src/segy.c
ajaust: Awesome!

test-data/README.md
```diff
@@ -25,6 +25,8 @@ Due to changes in the scripts after the files were originally created, reproduci
 | small-lsb.sgy                  | small.sgy converted to little endian                                        | ./flip-endianness --samples 50 small.sgy small-ps.sgy |                                            |
 | small-ps.sgy                   | Pre-stack file with 2 offsets.                                              | python make-ps-file.py small-ps.sgy  10 1 5 1 4 1 3   |                                            |
 | small.sgy                      | Inline-sorted file with 50 samples, 5 inlines-xlines test file.             | make-file.py small.sgy 50 1 6 20 25                   | Basic file used in testing                 |
+| stanzas-known-count.sgy        | Contains 3 extended text headers. Count is given in the binary header.      | manual                                                | Mix of ASCII and EBCDIC encodings.         |
+| stanzas-unknown-count.sgy      | Contains 3 extended text headers. Count is unknown, EndText is present.     | manual                                                | 2nd stanza is continuation of 1st          |
```
ajaust: It says "Contains 3 extended text headers". Is the number of stanzas and text headers identical in these cases? I guess that would mean that each stanza is fully contained within one extended text header.

In doubt, I we could update the description to "Contains 3 stanzas".

python/segyio/segyio.cpp
ajaust: Then I skip. 😆

python/segyio/segyio.cpp
ajaust: Make sense. Thanks for the explanation. 🙂

python/segyio/utils.py
ajaust: Yes, you are right. We would still need the initialiser so it would shorten the code.

I was thinking about it since this class does not contain any (real) logic but only data. I think you would get `__repr__` for free when using a dataclass (with type hints?).

I just leave this hear for the visual impression. I was curious how it would look like:
```python
from dataclasses import dataclass

@dataclass
class TraceHeaderLayoutEntry:
    name: str
    byte: int
    type: str # This might be the wrong type
    requires_nonzero_value: bool

    def __init__(self, name, byte, type, requires_nonzero_value):
        self.name = name
        self.byte = int(byte)
        self.type = type
        self.requires_nonzero_value = bool(int(requires_nonzero_value))

```

lib/src/segy.c
```diff
@@ -3027,8 +3027,8 @@ int segy_rotation_cw( segy_datasource* ds,
 /** Parses stanza header ((stanza_name)) from the text.
  *
  * If header is found, stanza name and stanza name length is extracted and set.
- * Encoding is determined based on the name. stanza_name memory is allocated on
- * heap. If allocation fails, SEGY_MEMORY_ERROR is returned.
+ * stanza_name memory is allocated on heap. If allocation fails,
```
ajaust: Just to be sure: It is the user's responsibility to free this memory. Is this correct? I see that we free it in the Python bindings but not otherwise.

ajaust (commented):
Thanks for the replies and explanations. To me it looks good overall. 🙂 

I marked most comments as resolved. The comments I did not mark as resolved usually had a (positive?) reaction, but I either didn't see any "Outdated" marker, or didn't find any code change related to the comment (I might have missed it). For some stuff, it could also be that the comment is not relevant anymore since the code disappeared due to moving to reading stanzas instead of extended textual headers.

I think I also added new 2-3 minor comments. I am curious

lib/src/segy.c
achaikou: Added info here too!

lib/src/segy.c
achaikou: Discussed, we leave it as is because the mess is in the standard anyway :smile:

achaikou (commented):
Thanks!
Something is fixed.
Everything is rebased and split back into meaningful commits again.

lib/src/segy.c
```diff
@@ -3033,3 +3033,142 @@ int segy_rotation_cw( segy_datasource* ds,
...
+) {
+    // start char should be (, end char should be )
+    const int ascii_start = 40;
+    const int ascii_end = 41;
+    const int ebcdic_start = 77;
+    const int ebcdic_end = 93;
+
+    char end_char;
+    int encoding;
+    if( text_start[0] == ascii_start && text_start[1] == ascii_start ) {
```
yngve793, line 3054: I guess we can keep this check even if we assume no ebcdic. Maybe it is required for standard text header?

lib/src/segy.c
```diff
@@ -3033,3 +3033,142 @@ int segy_rotation_cw( segy_datasource* ds,
...
+        *stanza_name_length = 0;
+        *stanza_name = NULL;
+        return SEGY_OK;
+    }
+
+    text_start += 2; // skipping ((
+
+    const char* search_start = text_start;
+
+    while( true ) {
```
yngve793: I would suggest to split this into two operations.
1: Find the "))"
2: Extract the stanza name and length. This can be done outside the loop.

PS: It seams like there is a unnecessary loop here.  Outer loop is the while loop. Inner loop is the "memchr(" search. Consider just to loop through the characters and check for ")"

lib/src/segy.c
```diff
@@ -3033,3 +3033,142 @@ int segy_rotation_cw( segy_datasource* ds,
...
+        memcpy( *stanza_name, text_start, *stanza_name_length );
+        if( encoding == SEGY_EBCDIC ) {
+            encode( *stanza_name, *stanza_name, e2a, *stanza_name_length );
+        }
+        return SEGY_OK;
+    }
+}
+
+/** Reads (parts of) text header from datasource and parses stanza header. */
+static int read_stanza_header(
```
yngve793, line 3096: Consider to combine `read_stanza_header` and `segy_read_stanza_header` into `segy_read_stanza_header`.
Part of `read_stanza_header` is the same for both cases used in `segy_read_stanza_header`.

python/segyio/segyio.cpp
yngve793: Resolved by data_size() function.

yngve793 (approved):
I am basically fine with it now. 
Have a look at the comments and adapt if you agree.

lib/src/segy.c
achaikou: As far as I understand, content itself is likely no-ebcdic (or we don't care if it is), but the stanza names still could be EBCDIC. At least this is current understanding of line

> Different documents in EBCDIC and ASCII may be mixed into a single SEG-Y file,
> there is no requirement they should be all EBCDIC or all ASCII

So this check is still valid.

Standard text header is not related to stanzas and does not start with `((` so check is irrelevant for it.

lib/src/segy.c
achaikou: The part that is the same is only header and offset definition. All other lines must be repeated.

I tried what you suggest and I still feel it is more readable to have logic extracted into a local `read_stanza_header` function.
I would be even more in favor of dropping `initial_read_size` optimization than combining those into one function.

Left as is.

lib/src/segy.c
achaikou: You are right! That is so much simpler :heart_eyes_cat:

achaikou (commented):
Thanks for the review!

ajaust (approved):
Just went through everything again to look for unresolved comments. Looks all good to me. 🙂

PR #643, opened 2025-09-24
Add valgrind job for python tests

achaikou:


.github/workflows/analyzers.yaml
```diff
@@ -105,3 +105,44 @@ jobs:
...
+      - name: Build
+        run: |
+          cmake --build build
+
+      - name: Run valgrind memcheck on default pytests
+        run: |
+          cd python
+          # test_no_16bit_overflow_tracecount is too large for valgrind, so skip
+          # only "invalid memory access" checks. "Memory leaks" would be caused by python anyway.
+          valgrind --tool=memcheck --error-exitcode=1 pytest -rPV -k 'not test_no_16bit_overflow_tracecount'
```
ajaust, line 148: In the other valgrind checks (line 94 - 107) the additional options `--leak-check=full --show-leak-kinds=all` are used. Is there any reason for deviating from this for the Python tests?

achaikou: Because for c++ we do not expect any leaks and we get
`All heap blocks were freed -- no leaks are possible`

For python, without `leak-check=full` option, we get only summary
```
==4714== LEAK SUMMARY:
==4714==    definitely lost: 37,279 bytes in 19 blocks
==4714==    indirectly lost: 83,383 bytes in 169 blocks
==4714==      possibly lost: 259,797 bytes in 123 blocks
==4714==    still reachable: 1,934,448 bytes in 1,605 blocks
==4714==         suppressed: 0 bytes in 0 blocks
==4714== Rerun with --leak-check=full to see details of leaked memory
```

That is a lot and it is likely caused by python. When I rerun with full check and "show leaks of all kinds", it prints around 10000 lines of leaks in stuff like "PyMem_RawMalloc" from python internal files.
Suppressing all of that to make sure we do not leak somewhere else doesn't seem like a good idea.

So printing all python "leaks" with `--leak-check=full` makes no sense.
Trace for the "invalid memory access errors" that we aim to find would be printed anyway.

ajaust (approved):

yngve793 (approved):
Looks good. No comment.

PR #644, opened 2025-09-29
WIP: preparations for introduction on n-traceheaders

achaikou:
Objectives:

- Figure out if we want to continue providing structural parameters to all segyio functions as parameters (or one struct parameter) or if we want to put them into `segy_datasource` struct
- If we want to put them into `segy_datasource`, how to best assure that `segy_datasource` always has all its properties set before main library functions are called
- If we want to use one initializing function, how to assure we know what property failed on error, so we still throw the same exception as before (unless we want to drop current exception granularity).
- Is there a totally different approach to all of this that I don't see?

### DIFF
--- a/applications/segyinfo.c
+++ b/applications/segyinfo.c
@@ -51,23 +51,24 @@ int main(int argc, char* argv[]) {
         exit( err );
     }
 
-    const int format = segy_format( header );
-    const int samples = segy_samples( header );
-    const long trace0 = segy_trace0( header );
-    const int trace_bsize = segy_trace_bsize( samples );
+    int format, samples, trace_bsize, traces;
+    long trace0;
+    int err_by;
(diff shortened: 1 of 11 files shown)


yngve793 (commented):
Some initial thoughts: (Inputs for further discussions)

Alternative 1) This is the same approach that is already used in the project, thus it will fit the current style. Most likely the fastest approach, but as @achaikou points out complexity increases. 

Alternative 2) This can be done along an axis. 
- A) Only store the values, similar to using parameters. We know the sizes of stuff but it is required to calculate the seek position.  
- B) Create a map for the segy file. C-user only ask for position of trace "i" and all seek logic is handled in one location. 
- C) An intermediate step. Calculate intermediate steps like total trace header size, etc. "trace0" is a precursor for this. 

Alternative 3)  I do not see all implications her, so it is hard to have an opinion.

PR #645, opened 2025-10-13
Introduce metrics

achaikou:
Preparations for supporting `traceheader_count` > 1.

This parameter would have been added in every place where `trace0` and `trace_bsize` are used.
After this PR all 3 parameters can be removed from the functions and will be treated the same way as `elemsize` currently is - set values would be read from `segy_datasource`.

To make code more consistent, endianness and encoding would now be treated the same way as all other parameters. With relation to this deducing endianness from file was implemented too.


lib/include/segyio/segy.h
```diff
@@ -422,6 +422,10 @@ int segy_samples( const char* binheader );
  */
 int segy_sample_interval( segy_datasource*, float fallback , float* dt );
 
+/* exception: returned int is elemsize of the format, not an error code. -1
+means uknown format. */
```
ajaust:
```suggestion
means unknown format. */
```

lib/src/segy.c
```diff
@@ -699,7 +699,7 @@ static int memclose( segy_datasource* self ) {
     return SEGY_OK;
 }
 
-static int formatsize( int format ) {
+int segy_formatsize( int format ) {
```
ajaust, line 702: Was the intention here to define the method as `static` in the header or do we not want this function to be static anymore?

python/segyio/segyio.cpp
```diff
@@ -641,46 +641,28 @@ PyObject* segyopen( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
+       case SEGY_INVALID_ARGS:
+           return RuntimeError( "unable to count traces, "
+                                "no data traces past headers" );
 
-        case SEGY_INVALID_ARGS:
-            return RuntimeError( "unable to count traces, "
-                                 "no data traces past headers" );
+       case SEGY_TRACE_SIZE_MISMATCH:
+           return RuntimeError( "trace count inconsistent with file size, "
+                                "trace lengths possibly of non-uniform" );
```
```suggestion
                                "trace lengths possibly non-uniform" );
```
This would change an existing error message so not sure if this is a breaking change.

applications/segyio-catr.c
```diff
@@ -699,29 +699,35 @@ int main( int argc, char** argv ) {
...
+
+                long long f = buf;
                 if( opts.nonzero && !f ) continue;
 
                 /*
                  * convert cannot-be-negative values to unsigned int, as mandated by SEGY-Y
                  * rev2
                  */
-                switch (f) {
+                switch (fields[j]) {
```
ajaust, line 662: Was this wrong before, i.e., before it switched on the field value instead of the field offset?

applications/segyio-catr.c
```diff
@@ -699,29 +699,35 @@ int main( int argc, char** argv ) {
...
                     case SEGY_TR_SAMPLE_COUNT:
                         f = (int)((uint16_t)f);
                         break;
+                    case SEGY_TR_SEQ_LINE:
+                    case SEGY_TR_SEQ_FILE:
+                        f = (long)((uint32_t)buf);
```
ajaust: In line 714 it does
```c
f = (int)((uint16_t)f);
```

and here it uses `buf` on the right-hand side

```c
f = (long)((uint32_t)buf);
```
Is there a reason for this difference? I just try to understand if this is done on purpose.

python/segyio/segyio.cpp
```diff
@@ -546,13 +546,13 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
         return -1;
     }
 
-    int err = segy_set_endianness( ds, endianness );
-    if( err ) {
-        ValueError( "internal: error setting endianness, was %d", endianness );
-        return -1;
+    if (endianness != SEGY_LSB && endianness != SEGY_MSB) {
+        int err = segy_endianness(ds, &endianness);
+        if( err != SEGY_OK ) return err;
```
ajaust: Didn't we check in line 479ff that `endianness` is set to a valid value? In this case don't need to determine the endianness from the file.

```c

    if( endianness != SEGY_MSB && endianness != SEGY_LSB ) {
        ValueError( "endianness must be set to a valid value" );
        return -1;
    }
```

Is a check for `endianness != -1` missing in line 479?

ajaust: I guess this is needed due to the changes in the next commit. Can you confirm this?

lib/src/segy.c
```diff
@@ -1475,26 +1475,19 @@ int segy_endianness( segy_datasource* ds, int* endianness ) {
...
+    *encoding = SEGY_EBCDIC;
+    char c;
 
-    ds->encoding = encoding;
+    int err = ds->seek( ds, 0, SEEK_SET );
+    if( err != SEGY_OK ) return err;
+    err = ds->read( ds, &c, 1 );
+    if( err != SEGY_OK ) return err;
+
+    // fist symbol is supposed to be 'C', which ASCII code is 67
```
ajaust:
```suggestion
    // first symbol is supposed to be 'C', which ASCII code is 67
```

python/segyio/segyio.cpp
```diff
@@ -547,11 +542,11 @@ int init( segyfd* self, PyObject* args, PyObject* kwargs ) {
     }
     ds.ds->lsb = endianness;
 
-    int err = segy_set_encoding( ds, encoding );
-    if( err ) {
-        ValueError( "internal: error %d setting encoding, was %d", err, encoding );
-        return -1;
+    if( encoding != SEGY_EBCDIC && encoding != SEGY_ASCII ) {
+        int err = segy_encoding( ds, &encoding );
+        if( err != SEGY_OK ) encoding = SEGY_EBCDIC;
```
ajaust: Why is it safe to set ` encoding = SEGY_EBCDIC;` instead of returning the error code? I though `err != SEGY_OK` would indicate a serious issue with the file during investigation of the file.

lib/include/segyio/segy.h
```diff
@@ -261,6 +261,42 @@ typedef struct {
...
+
+    /* Byte size of each trace (samplecount * elemsize). */
+    int trace_bsize;
+
+    /* Number of traceheaders describing each trace. */
+    int traceheader_count;
+
+    /* Number of traces in the file. */
+    int tracecount;
+} segy_metrics;
```
ajaust: I started thinking if a different name than `metrics` would have any benefit. Is see that in the experimental header the method `trace_meta_fromfile` uses the `segy_metrics` to describe `meta` (`data`?). In this sense, maybe metadata could be a sensible name as well.

I don't have a strong opinion (yet) and must think a bit more about it.

lib/src/segy.c
```diff
@@ -952,6 +975,64 @@ int segy_close( segy_datasource* ds ) {
...
+    int endianness,
+    int encoding,
+    int ext_textheader_count
+) {
+    if( endianness != SEGY_LSB && endianness != SEGY_MSB ) {
+        int err = segy_endianness( ds, &endianness );
+        if( err != SEGY_OK ) return err;
+    }
+    ds->metrics.endianness = endianness;
+    ds->lsb = endianness;
```
ajaust: ~Are we planning to remove the duplication of information within the `datasource` struct in the future?~

~From what I see, `endianness`, `encoding`, and `elemsize` appear to be stored within the `metrics` member of the `datasource` and directly within the `datasource`.~
I just found some interesting commits after the current one. 🙂

test-data/README.md
```diff
@@ -12,17 +12,19 @@ Due to changes in the scripts after the files were originally created, reproduci
...
 | long.sgy                       | Declares 60000 samples (value is written in 16-bit field in binary header). |                                                       | broken as proper trace headers are missing |
 | multi-text.sgy                 | Contains 4 extended text headers                                            | python make-multiple-text.py multi-text.sgy           | 1 sample, 1 trace                          |
 | shot-gather.sgy                | Some traces have common field record [2 3 5 8]/geophone group [1 2].        | make-shot-gather.py shot-gather.sgy                   | First trace value is field record value.   |
-| small-lsb.sgy                  | small.sgy converted to little endian                                        | ./flip-endianness --samples 50 small.sgy small-ps.sgy |                                            |
+| small-lsb.sgy                  | small.sgy converted to little endian                                        | ./flip-endianness --samples 50 small.sgy out.sgy      |                                            |
```
ajaust, line 28:
```suggestion
| small-lsb.sgy                  | small.sgy converted to little endian                                        | ./flip-endianness --samples 50 small.sgy small-lsb.sgy|                                            | 
```

ajaust (commented):
This looks great overall. 🎉 

I just left some minor comments for clarification. The only (potentially) major thing is the naming, i.e., "Is there a better name than `metrics`" for the properties of the file. I have to think a bit more about this since I have no answer myself.

lib/include/segyio/segy.h
achaikou: Interesting that my spellchecker seems to know the word `uknown`... :thinking:

lib/src/segy.c
achaikou: We want to remove `static`. We now make the function public, so available outside of the segy.c.

> static functions are functions that are only visible to other functions in the same file

[source](https://stackoverflow.com/a/558132)

python/segyio/segyio.cpp
achaikou: Will not fix because I remove this error message altogether in a later commit.
Those error messages help very little anyway.

applications/segyio-catr.c
achaikou: Yes.
Probably worked because there are no tests where sample count is near uint16_t max.

As those applications would not support revision 2.1 very well anyway, not bothering to test additionally.

applications/segyio-catr.c
achaikou: This was done on purpose by me trying to make one failing test pass :smile: 
I was satisfied with the first line that worked :smile: 

Apparently I can simplify that a bit. Though that CI might complain if I don't do the final cast, but it does not.

And for completeness I will change `f` to `buf` in `uint16_t` case as point of 'buf' is to keep the bytes and point of 'f' is to keep correctly interpreted value.

python/segyio/segyio.cpp
```diff
@@ -641,46 +641,28 @@ PyObject* segyopen( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
-        case SEGY_UNSIGNED_INTEGER_4_BYTE:      elemsize = 4; break;
-        case SEGY_UNSIGNED_SHORT_2_BYTE:        elemsize = 2; break;
-        case SEGY_UNSIGNED_INTEGER_8_BYTE:      elemsize = 8; break;
-
-        case SEGY_NOT_IN_USE_1:
-        case SEGY_NOT_IN_USE_2:
-        default:
-            break;
-    }
+   err = segy_traces( ds, &tracecount, trace0, trace_bsize );
```
yngve793: Less redefinition, warms my heart.

python/segyio/segyio.cpp
```diff
@@ -755,42 +737,8 @@ PyObject* segycreate( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
-        case SEGY_SIGNED_CHAR_1_BYTE:
-        case SEGY_UNSIGNED_CHAR_1_BYTE:
-            elemsize = 1;
-            break;
-
-        default:
-            assert(false || "format should already be checked" );
-            break;
-    }
+    segy_set_format( ds, format );
```
yngve793: Even less repetition :-)

test-data/README.md
achaikou: :smile: 
I did that on purpose because adding correct names would mean increasing line length even more. I am not sure I want to do that for "recreation" column, given that I found an easier solution :smile: 

But will do if current approach does not sit right with you! :+1:

python/segyio/segyio.cpp
achaikou: Yes.
You are right that this makes more sense together with the next commits :slightly_smiling_face:

Intention in the PR was to "read endianness if user's supplied one is not a valid option", but splitting it into easier-to-understand commits leads to some misunderstandings in other places :smile:

python/segyio/segyio.cpp
achaikou: 1. Original idea was that no matter what, file is EBCDIC encoded unless stated otherwise. So it is kind of fine.
2. Next commit moves this code to another place and final version lies in `segy_collect_metrics` function and in both places error is returned because I generally agree with you :smile: 
3. I think I simply missed it during development :slightly_smiling_face:   I can fix it, if you want, but only during rebase, as this requires an intrusive fixup because this code is removed already in the next commit :)

lib/include/segyio/segy.h
achaikou: :+1: 
Think and let me know. I am not able to make this decision on my own :smile: 

Current decision is based on `metrics` function in segyio.cpp that returns all those values. Also name is heavily used in python (for both this metrics and cube metrics).

"Meta" (or metadata) makes some sense too and probably slightly more on point.

achaikou (commented):
Thanks for the review!

lib/src/segy.c
```diff
@@ -1371,59 +1371,35 @@ int segy_field_forall( segy_datasource* ds,
...
-        SEGY_BIN_TRACE_FLAG,
-        SEGY_BIN_EXT_HEADERS,
-    };
-
-    const int bytes2_len = sizeof( bytes2 ) / sizeof( int );
-
-    for( int i = 0; i < bytes2_len; ++i ) {
-        const int offset = bytes2[ i ] - ( SEGY_TEXT_HEADER_SIZE + 1 );
-        bswap16_mem( xs + offset, xs + offset );
+static int bswap_bin( const segy_datasource* ds, char* xs ) {
```
yngve793, line 1447: A much cleaner solution.

lib/include/segyio/segy.h
```diff
@@ -443,16 +443,12 @@ int segy_format( const char* binheader );
  */
 int segy_set_format( segy_datasource*, int format );
 
-/* set file as LSB/MSB (little/big endian)
+/* Find out whether file is LSB/MSB (little/big endian).
```
yngve793: Insert new line at line 431. After the definition of `int segy_format( const char* binheader );`

lib/src/segy.c
```diff
@@ -1449,16 +1449,29 @@ int segy_set_format( segy_datasource* ds, int format ) {
...
-        case SEGY_LSB:
-        case SEGY_MSB:
+int segy_endianness( segy_datasource* ds, int* endianness ) {
+    char binheader[SEGY_BINARY_HEADER_SIZE];
+    // by default read as msb
+    int err = segy_binheader( ds, binheader );
+    if( err != SEGY_OK ) return err;
+
+    segy_field_data fd;
+    segy_get_binfield( binheader, SEGY_BIN_INTEGER_CONSTANT, &fd );
```
yngve793, line 1524: Happy to see SEGY_BIN_INTEGER_CONSTANT getting used.

python/segyio/create.py
```diff
@@ -6,6 +6,8 @@
 from .utils import (
     FileDatasourceDescriptor,
     StreamDatasourceDescriptor,
+    c_endianness,
```
yngve793: The `c_` prefix here seams a bit strange. Are there other endianness we can confuse it with.  Endianness sounds like a data stream property and not a "c" property. 

Same for encoding.

applications/segyio-catr.c
```diff
@@ -470,7 +470,6 @@ static struct options parse_options( int argc, char** argv ){
     static struct option long_options[] = {
         { "trace",          required_argument,  0, 't' },
         { "range",          required_argument,  0, 'r' },
-        { "format",         required_argument,  0, 'f' },
```
yngve793, line 473: Initially I thought that this probably reduced usability of the library. In real use especially with old data files I can imagine that this is a very useful feature. On the other hand newer files will probably not have the problem. In work case an old version can be used to read the data.

lib/include/segyio/segy.h
yngve793: Metadata or meta something is a commonly used name for this.

lib/include/segyio/segy.h
```diff
@@ -261,6 +261,42 @@ typedef struct {
     segy_entry_definition offset_to_entry_definion[SEGY_TRACE_HEADER_SIZE];
 } segy_header_mapping;
 
+typedef struct {
```
yngve793, line 268: This is truly a significant structural improvement. Identifying and keeping track of the state defining values has been a hassle in this project. Now that all these are gathered in one place and much more important accessed through segy_metrics it is so much easier to keep control. For instance writing to any segy_metrics value will have global implications. Previous it was a research job to identify the scope and how widely a variable was used.

yngve793 (approved):
Very nice solution. :-)

lib/include/segyio/segy.h
achaikou: Not done because I remove `segy_set_format` completely in one of the next commits.

python/segyio/create.py
achaikou: Was not introduced in this commit but earlier.

There are two representations of "endianness" and "encoding" properties - one in Python, another in C.
`c_endianness` and `c_encoding` functions are responsible to translate value for encoding/endianness in python ("ebcdic", "little") to corresponding value in C (0 or 1).
I think I named them so because `python_endianness_to_c_endianness` sounded too long.

Could change the name if you suggest what to change it to. :slightly_smiling_face:

applications/segyio-catr.c
achaikou: This whole change is based on PO being unaware of existing files with wrong format value set.
In the issue where this code was introduced there was also no talk about this.

So I too hope this is really not a valid case anymore and as you say, users could use previous versions, if they have to  :+1:

lib/include/segyio/segy.h
achaikou: I decided that you guys were right, if we already agreed to change it to metadata or something I can as well do it now. So I changed everything I introduced to "metadata".
Don't really have a desire to hunt for existing occurrences of metrics, so that one is left for future :smile:

achaikou (commented):
:heart:

python/segyio/create.py
yngve793: Both `c_endianness` and `c_encoding` preforms a mapping from text to an int value. Neither of the functions are using any part of the "C" code directly, but they translate the open parameters into parameters that can be understood by the c-code. My suggestion is to name them in accordance with the function they preform. I.e. "map_endianness" and "map_encoding" or something similar.

yngve793 (approved):
Looks good.

lib/include/segyio/segy.h
ajaust: Maybe it was added as (permanent) exception to the dictionary at some point?

test-data/README.md
ajaust: It is not that important to me that is must be changed. 🙂

python/segyio/segyio.cpp
ajaust: Keep it as is since it is changed in the next commit anyway. 👍

lib/include/segyio/segy.h
ajaust: I am happy with the fixup. I will leave this comment as "unresolved" so Yngve can also confirm that he is happy as well.

ajaust (approved):
The changes look good to me. The fixups are asle nice. 👍 

All my comments are resolved. I just left one open for reference since Yngve commented on the same thing (renaming `metrics` to `metadata`).

lib/include/segyio/segy.h
yngve793: Happy :-)

python/segyio/create.py
yngve793: Fixed.

PR #646, opened 2025-10-14
Build NumPy from source in benchmarks

ajaust:
Just testing performance.


ajaust:
I will close this for now because building NumPy from source does not seem to have any influence on the benchmark results.

PR #647, opened 2025-10-15
Silent warnings emitted by Python binding's build process

ajaust:
This addresses failing builds of the Python bindings if one enables additional warnings and turns warnings into errors.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -10,7 +10,7 @@ on:
 jobs:
   test:
     runs-on: ${{ matrix.os }}
-    name: Build and test ${{ matrix.build_type }} on ${{ matrix.os }} ${{ matrix.compiler }} ${{ matrix.cmake_generator }}
+    name: Build and test ${{ matrix.build_type }} on ${{ matrix.os }} ${{ matrix.compiler }} ${{ matrix.cmake_generator }} BUILD_PYTHON=${{ matrix.build_python }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -162,3 +162,24 @@ jobs:
         run: |
           cd python
           pytest -rP --datasource ${{ matrix.stream }} test/stream.py
+
+  python-all-warnings:
```
achaikou: Any particular reason why you added it as a separate job instead of having  `-DBUILD_PYTHON` parametrized in the `test` job?

I am not saying it would be better because you would need to add `python = on/off` parameter (and I recall duplicating it in each step was the best option), but just wondering :slightly_smiling_face:

ajaust: This had no particular reason. It felt simpler (I always have to do some debugging when fiddling with GitHub's matrix builds), but it has the disadvantage that only one compiler is tested. We use at least two compilers during development (GCC on Linux, Clang on MacOS) so testing both (and thus using the matrix build) would make sense.

I have some more commits in another branch (see [PR in my fork](https://github.com/ajaust/segyio/pull/8/commits)) that we can move into the PR here. The updated matrix build does not look as bad I expected and it is simple to test for warnings on all platforms.

What do you think?

achaikou: I like the fixup proposition :+1: 
It is more code, but we keep similar things together and as you say, test two compilers.

So go for what you like more :)

Upd: Note that you say

> We use at least two compilers during development (GCC on Linux, Clang on MacOS)

and yet you added `clang + ubuntu`, not `macos (with clang by default)` case.

ajaust: I moved the fixup into this PR. I also added another fixup to replace clang + Ubuntu by clang + MacOS (good catch!). This also required the use the GitHub Action for setting up Python. Otherwise, there was an issue on Mac that the Python dependencies could not be installed as `pip` tried to install them into the system-managed environment.

In the PR in my fork, I also experimented with a matrix entry for Windows, but this gives a linker issue during linking the Python binding's library. I don't think I can easily debug this without a Windows machine so I did not move this commit into the PR here. I am not sure if we want to spend energy of fixing it at the moment.

achaikou: I agree - don't bother with Windows.
I think it looks good and can be rebased and merged :tada:

achaikou: Oh, and the error you see on Windows is about
`segy_memopen`
missing in segy.def file.
Good point, it should be there, I think.

ajaust: I just rebased and squashed everything. Waiting now that the tests pass. :) 

Good to hear that the error message on Windows is reasonable and simple to fix.

PR #648, opened 2025-10-17
Fix GitHub security warnings

ajaust:
This should tackle all current warnings. Permissions are set as low as possible. I did not test the Action for publishing to PyPI as this needs a tagged release.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -1,4 +1,5 @@
 name: Wheels
+permissions: {}
 
 on:
   push:
--- a/.github/workflows/analyzers.yaml
+++ b/.github/workflows/analyzers.yaml
@@ -1,4 +1,5 @@
 name: Run analyzers
+permissions: {}
 
 on:
   push:
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -1,4 +1,5 @@
 name: Benchmarks
+permissions: {}
 
 on:
   push:
--- a/.github/workflows/bigendian.yaml
+++ b/.github/workflows/bigendian.yaml
@@ -1,4 +1,5 @@
 name: Test bigendian
+permissions: {}
 
 on:
   push:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,4 +1,5 @@
 name: Build and Test
+permissions: {}
 
 on:
   push:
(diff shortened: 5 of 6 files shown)


.github/workflows/wheels.yaml
```diff
@@ -65,6 +66,8 @@ jobs:
           path: ./wheelhouse/*.whl
 
   publish:
+    permissions:
+      secrets: read
```
achaikou: I suspect it might be incorrect. I don't find "secrets" in permissions list and get `Unexpected value 'secrets'` from my yaml-schema checking.
(and it actually failed on [my fork](https://github.com/achaikou/segyio/actions/runs/18657579587), but here for some reason workflow wasn't started and nothing even complained)

I think in si4ti we enabled "trusted publisher". Here probably we didn't, but I recall an email that screamed at me that we should.
Then we would need `id-token: write` permission.

So the only way to test what is going on it is to create a segyio repo on test pypi and change all properties for testing to point to `repository-url: https://test.pypi.org/legacy/`

.github/actions/benchmark/action.yaml
```diff
@@ -1,5 +1,6 @@
 name: Benchmark
 description: benchmark execution
+permissions: {}
```
achaikou: Looks like this property is not allowed, because it is action, not workflow.

ajaust: That is a good point. I will drop this. I think GitHub also never complained about the permissions inside this file.

.github/workflows/wheels.yaml
ajaust: I will test it out on TestPyPI. I still have an account for TestPyPI from when working on si4ti. The `secrets` are at least listed as repository permissions [in this part of the docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens#account-permissions). That is where I got the keyword from.

We could consider setting up the trusted publisher as this is the recommended way to go for if you think that is about time for that.

ajaust: I added a fixup to drop the permission. The linter and you are right. The `secrets` permissions seems to apply to another token than the one created inside the workflow. 

Due to me adding this non-existing permission the whole workflow broke. This caused the workflow to be skipped. Removing the permission made the workflow for wheels run again. 

Currently, only the wheels for Ubuntu ARM are built. I wanted to reduce the amount of time I need to wait when publishing stuff to PyPI.

ajaust:
It looks like the tests (also the large ones) pass now. I think the error we observed on the Windows tests was due to me renaming the package (inconsistently) from `segyio` to something else. There seems to be a step in the Windows build that requires the package to be named `segyio`.

PR #650, opened 2025-10-17
Open file with shifted layout

achaikou:
With this PR we can open files where trace fields are located not at their default position.

We cannot yet open our available 2.1 file, cannot correctly read fields and do not support non-standard datatypes.
Everythin related to that would happen in the next PRs.


python/segyio/trace.py
```diff
@@ -779,13 +779,36 @@ class Attributes(Sequence):
...
+            "reeltrc": np.uint32,
+            "linetrc8": np.uint64,
+            "reeltrc8": np.uint64,
+            "coor4":    np.int32,
+            "elev4":    np.int32,
+            "time2":    np.int16,
+            "spnum4":   np.int32,
+            # "scale6" unspported, unclear what to do yet
+        }
+        entry = next(e for e in traceheader_layout if e.byte == field)
```
achaikou: Benchmarks are likely failing because of this line.

This is not optimized to search by `.byte` or by `.name` because I decided that these operations are not executed in the loop, so it shouldn't matter much for the user. Besides it is not yet clear what operations would be most frequent - search by name or by byte.

However benchmark tests do execute `.attributes` in a loop and thus clearly see performance degradation. There would be no easy way to not failing this test besides dropping it.

So it seems like easiest solution is to optimize: introduce `TraceHeaderLayout` class instead of List of `TraceHeaderLayoutEntry` where we could re-map our list by name or by index.

achaikou: Unfortunately situation is more complicated than that.  `open` is also inside the loop, so number of `.attribute` calls is still the same for each iteration

If I split benchmark tests into testing .open and .execute separately (so that only function part itself gets executed in the loop), results are confusing.
Suddenly test that ran in 600ms becomes 4ms for open and 60ms for execute. `.attributes` and `.depth` mmap calls consistently go over threshold up to 100%.
I tried to removed all other differences and result is inconsistent. Some slowness is random, some - likely not.
Good news is that those times are still fast enough not to make that much difference.

So in this PR I didn't split anything and lets hope that some changes that I made would prevent tests from failing too often :smile:

python/segyio/utils.py
```diff
@@ -146,6 +146,25 @@ def __repr__(self):
         return msg.format(self.name, self.byte, self.type, self.requires_nonzero_value)
 
 
+class TraceHeaderLayout:
+    def __init__(self, entries):
+        self.entries = list(entries)
+        self.by_byte = {entry.byte: entry for entry in self.entries}
+        self.by_name = {entry.name: entry for entry in self.entries}
```
ajaust: Should the members be marked as private, i.e, prefix `entries, `by_byte`, and `by_name` with `_`?

python/segyio/utils.py
```diff
@@ -146,6 +146,25 @@ def __repr__(self):
...
+
+    def get_by_byte(self, byte):
+        return self.by_byte.get(byte)
+
+    def get_by_name(self, name):
+        return self.by_name.get(name)
```
ajaust: Depending on how we want to access/work with the members of the class, we could drop the `get_` methods and directly access the members similar to the approach in `TraceHeaderLayoutEntry`.

python/test/segyio_c.py
```diff
@@ -181,6 +181,19 @@ def test_custom_iline_xline():
...
+    f = segyfile(str(testdata / 'small.sgy'), "r", 0).segyopen(iline=9, xline = 17)
+    layout = f.traceheader_layout()
+
+    standard_header_layout = layout["SEG00000"]
+    entry = standard_header_layout.get_by_byte(9)
+    assert entry.name == "iline"
+
+    xline_entries = [e for e in standard_header_layout if e.name == "xline"]
+    # entry names do not get overwritten if new iline/xline are defined
+    assert len(xline_entries) == 2
```
ajaust, line 193: By initially looking at this I am slightly confused why we would have two entries. Is the "old" definition of the `xline` header not replaced by the new xline definition when opening the file with `xline = 17`?

python/segyio/segyio.cpp
```diff
@@ -2379,7 +2379,16 @@ int overwrite_inline_xline(
         }
         uint8_t il = static_cast<uint8_t>( iline );
         mapping.name_to_offset[SEGY_TR_INLINE] = il;
-        mapping.offset_to_entry_definition[il - 1] = { SEGY_ENTRY_TYPE_INT4, false, NULL };
+
```
yngve793: - Consider splitting overwrite_inline_xline into two functions. 
- Consider making one generic function as the code is very similar.

python/segyio/segyio.cpp
```diff
@@ -2772,6 +2780,127 @@ extern "C" int segy_parse_layout_xml(
...
+) {
+    size_t entry_type_map_size = sizeof( entry_type_map ) / sizeof( entry_type_map[0] );
+
+    std::string entry_type;
+    for( size_t i = 0; i < entry_type_map_size; ++i ) {
+        if( def.entry_type == entry_type_map[i].segyio_entry_type ) {
+            entry_type = entry_type_map[i].spec_entry_type;
+            break;
+        }
+    }
```
ajaust, line 2761: If this is actually too expensive, search for every entry type, I guess the simplest is to set up some lookup table/method to avoid the search.

python/segyio/segyio.cpp
```diff
@@ -2772,6 +2780,127 @@ extern "C" int segy_parse_layout_xml(
...
+    if( !trace_header_layout ) {
+        Py_DECREF( header_name );
+        Py_DECREF( entries );
+        return SEGY_INVALID_ARGS;
+    }
+
+    PyDict_SetItem( layout_dict, header_name, trace_header_layout );
+    Py_DECREF( header_name );
+    Py_DECREF( entries );
+    Py_DECREF( trace_header_layout );
```
ajaust, line 2827: I just must ask because I don't have full control over Python's reference counting.

Why do we have to decrease the reference counter here? Did the creation of the `TraceHeaderLayout` object (line 2853) and adding it to the dictionary (line 2860) increase the counters instead of taking ownership over the objects (if this concept exists in the Python interpreter).

python/segyio/open.py
```diff
@@ -80,12 +80,16 @@ def open(filename, mode="r", iline = 189,
         File access mode, read-only ('r', default) or read-write ('r+')
 
     iline : int or segyio.TraceField
-        Inline number field in the trace headers. Defaults to 189 as per the
-        SEG-Y rev1 specification
+        Overrides inline field offset in the trace headers. If no value is set,
+        value defined in xml mapping or standard-defined 189 would be used. By
+        setting this value user takes over responsibility of assuring parsing
+        can be done correctly.
```
ajaust: I am just thinking if we should rephrase it slightly to emphasize the order of precedence (?) in which the value is chosen. I am just writing some ideas so that I can also think about them.

### Idea 1
```suggestion
        Overrides inline field offset in the trace headers. If no value is set,
        value defined in xml mapping is used and if this one is absent the 
        standard-defined value 189 is used. By setting this value user takes
        over responsibility of assuring parsing can be done correctly.
```

### Idea 2

```suggestion
        Overrides inline field offset in the trace headers. segyio looks up
        the field offset in the following order:
        
        1. User provided `iline` parameter in `open`
        2. Offset value defined in XML trace header mapping (see section D8 of
           SEGY rev. 2)
        3. Default value of 189
        
        By setting this value user takes over responsibility of assuring parsing
        can be done correctly.
```

python/segyio/open.py
```diff
@@ -80,12 +80,16 @@ def open(filename, mode="r", iline = 189,
...
 
     xline : int or segyio.TraceField
-        Crossline number field in the trace headers. Defaults to 193 as per the
-        SEG-Y rev1 specification
+        Overrides crossline field offset in the trace headers.  If no value is
+        set, value defined in xml mapping or standard-defined 193 would be used.
+        By setting this value user takes over responsibility of assuring parsing
+        can be done correctly.
```
ajaust: Just a reminder that if we update the description for the `iline` parameter, we should update this accordingly.

python/segyio/segyio.cpp
```diff
@@ -507,6 +507,12 @@ int set_traceheader_mappings(
  * SEGY_OK both when data was successfully extracted or not found. */
 int extract_layout_stanza( segyfd* self, std::vector<char>& layout_stanza_data );
 
+/**
+ * Returns the dictionary {header_name: [TraceHeaderLayoutEntry] of traceheader
+ * mappings.}
+ */
+PyObject* traceheader_layout( segyfd* self );
```
ajaust: I am contemplating if I would prefer 

```suggestion
PyObject* traceheader_layouts( segyfd* self );
```

over some renaming at another locations. I find it currently hard to distinguish inside the code if we work with one specific layout (e.g., `SEG00000`) or the dictionary of layours because these tend to be called `layout`.

Maybe working with `traceheader_layout` if one works with the dictionary of trace header layouts (`SEG00000` etc.) and `traceheader_entries` if one works with the layout defined for a specific trace header. `traceheader_layouts` and `traceheader_layout` could also work, but I am not sure if a single `s` could be easy to miss.

I am not sure yet though how a good solution could/should look like.

python/segyio/trace.py
```diff
@@ -779,12 +779,35 @@ class Attributes(Sequence):
...
+    }
+
+    def __init__(self, field, segyfd, traceheader_layout, tracecount):
         super(Attributes, self).__init__(tracecount)
         self.field = field
         self.segyfd = segyfd
         self.tracecount = tracecount
-        self.dtype = np.intc
+
+        entry = traceheader_layout.get_by_byte(field)
```
ajaust: I am thinking on how to make it easier for me to follow what we are working with by renaming things slightly:
```suggestion
        entry = traceheader_layout.entry_by_byte(field)
```

test-data/README.md
```diff
@@ -97,5 +97,6 @@ Files are created manually.
 | mapping-invalid-root.sgy       | Root does not define standard header layout.                  |
 | mapping-invalid-value.sgy      | Last standard header entry byte is out of range.              |
 | mapping-multiple-stanzas.sgy   | Other stanzas are present before and after layout stanza.     |
+| mapping-shifted.sgy            | Default mapping with all offsets shifted by 1 byte.           |
```
ajaust, line 103: Just for clarification. The mapping is embedded inside as XML and therefore we do not have to specify the offsets when opening the files inside the tests. Is this correct?

ajaust (commented):
Overall, I think the changes look good. I think my main concern/confusion is when working with the dictionary of layouts or a specific layout and the corresponding entries. I am not sure if we are able to make the code simpler to read if we would rename some variable. I made some suggestions, but I am not sure if there is a better solution. I am curious about your thoughts. 🙂

python/segyio/segyio.cpp
```diff
@@ -507,6 +507,12 @@ int set_traceheader_mappings(
  * SEGY_OK both when data was successfully extracted or not found. */
 int extract_layout_stanza( segyfd* self, std::vector<char>& layout_stanza_data );
 
+/**
+ * Returns the dictionary {header_name: [TraceHeaderLayoutEntry] of traceheader
```
achaikou: Fix the docs

python/segyio/segyio.cpp
achaikou: Hm. Good and difficult point.

I dislike all the distinctive names I can come up with.
- keeping `traceheader_layout` for all traceheaders using `traceheader_entries` to describe one traceheader also sounds confusing to me  and "entries" describes details, not purpose.
- going back to "mapping" word in any of them would be confusing as both layout and mapping kind of mean the same, but I percieve "layout" as higher-level name than "mapping".
- `traceheader_layouts` is the most fitting one, but as you say is easy to miss.
- `traceheader_layout_dict` name looks a bit weird together with other ones we use in bindings.
- `traceheader_layout_collection` is also out of place.
- `segy_layout`, by the name of the root entry, also looks stupid because we already know it is segy.

I think out of all these I would vote for `traceheader_layouts`. Out of context it should be quite visible if we are talking about one layout or several.

python/segyio/segyio.cpp
achaikou: For this one must look at the docs

> *PyObject_CallFunctionObjArgs
>     Return value: New reference.

New reference means increased reference counter.

> int PyDict_SetItem( *p, *key,  *val)
>  This function does not steal a reference to val.

"Does not steal" means that it increases reference counter for itself. It does not take ownership of the provided ref.

That means that code still has one reference (returned by `PyObject_CallFunctionObjArgs`) that is not taken care of by anyone. So we must delete it.

python/segyio/utils.py
achaikou: `entries` - probably no, I am fine with user accessing the list if they need it. by_byte and by_name - good point.

python/segyio/utils.py
achaikou: I think here functions are more natural, as by_name and by_byte are internal performance-optimized representations of the same entity, but can be discussed :)

python/test/segyio_c.py
achaikou: You are correct, it is not replaced.

I had no idea what to do with old entry if a new one is given. I decided to let it be, assuming that if you override iline-xline with mapping present, you deserve to be confused :smile: 

Should original iline/xline be searched by name and its entry fully deleted, as if it never existed? I don't know.

python/segyio/trace.py
achaikou: Works for me!

test-data/README.md
achaikou: Mapping is embedded inside as XML. All values are shifted.

> we do not have to specify the offsets when opening the files inside the tests

It is not even an option to specify offsets when opening files. It is valid for inline-xline only.

python/segyio/segyio.cpp
achaikou: Hm... Maybe you are right, maybe that also contributes to performance degradation.

I added lookup maps for types. Maybe if we have time too implement extension 1 header one, I will add lookup maps for names then too.

python/segyio/open.py
achaikou: Chose first option because this is docs for one parameter in already large function documentation. Would prefer to keep it on the small side.

achaikou (commented):
Thanks for the review!
Something is updated, something will be updated after we agree upon it :slightly_smiling_face:

python/segyio/segyio.cpp
ajaust: Great, consider this point as resolved. 👍 

Out of curiosity: Did you see any (obvious) difference in the performance benchmarks? Is the test that was failing occasionally (due to this PR) passing more consistently now?

python/segyio/utils.py
ajaust: Nah, I don't require a change here. If it is consistent with the general interface, leave it as it is. 🙂

python/segyio/segyio.cpp
```diff
@@ -2402,9 +2419,31 @@ int initialize_traceheader_mappings(
...
-        );
+        segy_header_mapping mapping = ds->traceheader_mapping_standard;
+
+        /* Entry names provided by xml are of unknown length, so we must
+         * allocate them on heap and delete them afterwards. Default C maps have
+         * their names set to NULL. If we provided default names for entries
+         * inside of the C code, then:
+         * - if we allocated them on heap, C library users would have to
+         *   deallocate the memory and for now we don't have good enough reason
+         *   to force users to call some dealloc function.
```
achaikou: Looks like I am lying. We have `segy_close` that should be called if `segy_open` was.

So maybe we can and should allocate on heap everything we want to!

python/test/segyio_c.py
ajaust: I also did not think too much about this before and I also see reasons to leave the behavior as is. I could also see arguments to change it and remove the old `iline`/`xline` definitions. In this sense I am also uncertain.

The behavior was not introduced in this PR either (but in `67b8c41`?) so we do not need to change it here. Maybe we get stronger feelings about this later. 🙂

python/segyio/segyio.cpp
ajaust: I agree. It is hard. I also agree that `traceheader_layouts` is probably the best fit. If it would not be too much work, I would like to see how it looks like if one changes it to `traceheader_layouts`, but I won't insist on it. 🙂

python/segyio/segyio.cpp
ajaust: Do you have anything in mind that you would like to allocate on heap to simplify our lives or to improve the code?

ajaust (approved):
The changes look good. I don't have anything in mind that must be changed. 😄 

I am just curious to see the code using `layout`/`layouts` part to get an impression whether this looks better, *but only* if it is not too much work.

python/segyio/segyio.cpp
achaikou: I immediately thought about strings from this very commit "Assign names to default traceheader mappings".

I thought we could allocate them on heap already in C code and remove with `segy_close`.
But it probably would still be ugly in practice though, as
1) I can't just allocate the memory on heap in static maps
2) deletion anyway would happen only for `segy_open` - `segy_close` path, so to achieve something I will basically have to copy code from here to `segy_open`, and leaving it here is likely least confusing option.

So I don't know how I can in practice implement my enlightenment here :pouting_cat:
But I will at least change the text, so it is no longer lying.

---
Other thing that we might benefit from is mappings. The other PR showed that we need propitiatory header mappings in the C code too because endianness swap is supposed to be private to the library. So now I have to pass them as arguments.

But as we do not create those maps in C, seems like that is not going to happen too :pouting_cat: 

---

But I am happy anyway. My thinking recently was limited by "but there is no dealloc function in C, and we probably should not introduce one now". Now I am at least aware that `segy_close` can do the job :smile:

python/segyio/segyio.cpp
achaikou: No, it looks like slowness is probably not related to this PR after all :grin: 
My other PR failed on attributes today too...

python/segyio/segyio.cpp
achaikou: Done! :magic_wand:

python/test/segyio_c.py
```diff
@@ -182,7 +182,7 @@ def test_custom_iline_xline():
         segyfile(str(testdata / 'small.sgy'), "r", 0).segyopen(xline=241)
 
     f = segyfile(str(testdata / 'small.sgy'), "r", 0).segyopen(iline=9, xline = 17)
-    layout = f.traceheader_layout()
+    layout = f.traceheader_layouts()
 
     standard_header_layout = layout["SEG00000"]
```
ajaust: If we keep `layouts`, then maybe also update the temporary variable:
```suggestion
    layouts = f.traceheader_layouts()

    standard_header_layout = layouts["SEG00000"]

File: python/segyio/segyio.cpp (hunk 10 above)
Comment (ajaust) [reply python/segyio/segyio.cpp]: Great. I think I prefer the new version using `layouts`. 🙂

**Review by ajaust — APPROVED** [review]
Great! Thanks for addressing this so quickly. It looks good from my side. I had just one minor suggestion in the newly added fixups.

File: python/test/segyio_c.py (hunk 15 above)
Comment (achaikou) [reply python/test/segyio_c.py]: Done, but as I rebased, it is not a fixup :)

File: python/segyio/segy.py [hunk 16]
```diff
@@ -27,14 +27,15 @@ def __init__(
...
-        iline=189,
-        xline=193,
         endian='big'
     ):
 
         self._datasource_descriptor = datasource_descriptor
-        self._il = iline
-        self._xl = xline
+
+        self.traceheader_layouts = fd.traceheader_layouts()
```
achaikou: Maybe _this_ should be private though.
It is likely not the final way we want users to access the data, so maybe we should keep it kind-of-hidden.

python/segyio/segyio.cpp
```diff
@@ -2363,9 +2380,31 @@ int initialize_traceheader_mappings(
     segy_datasource* ds = self->ds;
 
     if( layout_stanza_data.empty() ) {
-        self->traceheader_mappings.insert(
-            self->traceheader_mappings.begin(), ds->traceheader_mapping_standard
-        );
+        segy_header_mapping mapping = ds->traceheader_mapping_standard;
+
+        /* Entry names provided by xml are of unknown length, so we must
```
yngve793, line 2398: Given the limitations I think that this is a reasonable approach.

python/segyio/utils.py
```diff
@@ -141,6 +141,25 @@ def __repr__(self):
         return msg.format(self.name, self.byte, self.type, self.requires_nonzero_value)
 
 
+class TraceHeaderLayout:
+    def __init__(self, entries):
+        self.entries = list(entries)
+        self._by_byte = {entry.byte: entry for entry in self.entries}
```
yngve793, line 147: Nice to give multiple access options.

python/segyio/segyio.cpp
```diff
@@ -2733,6 +2751,122 @@ extern "C" int segy_parse_layout_xml(
...
+    PyDict_SetItem( layout_dict, header_name, trace_header_layout );
+    Py_DECREF( header_name );
+    Py_DECREF( entries );
+    Py_DECREF( trace_header_layout );
+
+    return SEGY_OK;
+}
+
+PyObject* traceheader_layouts( segyfd* self ) {
+    PyObject* module = PyImport_ImportModule( "segyio.utils" );
```
yngve793: The word `module` is flagged as a keyword in Github and VScode. Consider changing the variable name slightly.

yngve793 (approved):
The PR is in my opinion fine. Added some minor comments and suggestions. None of them are critical or principally important.  

1) Setting names to NULL in C.  We have discussed this in multiple iterations and the limitations in C are what they are. All in all I consider the final solution to be a good choice given the available options. 

2) `segy_field_forall` not supporting float. This is fine for now, but do we have a clear idea on how to incorporate this without breaking to much of the existing code structure? If we have a rough plan then that is great. If not it would make sense to rank importance versus implications.

achaikou (commented):
Thanks for the review!!

2. This does not support floats yet because floats are not added to all switch-cases. So I think it should be dealt with in the next PR without any additional problems.

python/segyio/segyio.cpp
yngve793: Thanks, very nice.

python/segyio/segyio.cpp
yngve793: Nice, fix.

PR #651, opened 2025-10-27
Adding traceheaders

achaikou:
Supports opening files with more than just standard traceheader.
Data is available internally in python, but not for end users yet.
    
Full disclosure regarding matlab code changes: taken approach is "if CI job passes, we are good" :grin:


lib/src/segy.c
```diff
@@ -942,13 +942,20 @@ int segy_flush( segy_datasource* ds ) {
...
+    long long offset,
+    long trace0,
+    int trace_bsize
+) {
+    long long trace_size = trace_bsize +
+                           SEGY_TRACE_HEADER_SIZE * ds->metadata.traceheader_count;
+    long long pos = trace0 +
+                    trace * trace_size +
+                    traceheader * SEGY_TRACE_HEADER_SIZE +
+                    offset;
```
ajaust: Do we need to check if `traceheader` is smaller than the number of trace headers? I could see that we start seeking inside the trace samples without noticing, i.e. still within the file (=no error) but the wrong portion of it. 

Example:
The file has 3 trace headers per trace (standard trace header + 2 extended trace headers) and 1000 samples. Calling with `seek_trace_header = 3` would seek within for the header within the samples of the current trace instead of the header.

1. Could this ever happen?
2. Do we safeguard against this or is this considered a user error.

applications/segyio-catr.c
```diff
@@ -644,7 +642,7 @@ int main( int argc, char** argv ) {
                                      "out of range" ) );
             if( i > numtrh ) break;
 
-            err = segy_traceheader( src, i - 1, trheader, trace0, trace_bsize );
+            err = segy_traceheader( src, i - 1, trheader  );
```
ajaust:
```suggestion
            err = segy_traceheader( src, i - 1, trheader );
```

lib/include/segyio/segy.h
```diff
@@ -563,8 +563,20 @@ int segy_trace_bsize( int samples );
...
-/* byte-offset of the first trace header. */
-long segy_trace0( const char* binheader );
+
+/* Sets byte-offset of the first trace header.
+ *
+ * Returns the value from 3521–3528 if it is positive. Otherwise function would
+ * calculate implied trace0 consulting ext_textheader_count. If provided value
+ * is negative, attempt would be made to read value from fields 3505–3506 and
+ * error code is returned if this value is negative too.
+ */
```
ajaust, line 581: Is it in line with our current approach that we expect at least one of these number to be set? We could also start counting the number of text headers ourselves if all provided values are negative.

lib/src/segy.c
```diff
@@ -1559,6 +1563,16 @@ int segy_encoding( segy_datasource* ds, int* encoding ) {
     return SEGY_OK;
 }
 
+static int segy_revision( const char* binheader, int* revision ) {
```
ajaust, line 1616: ~I was just confused for a moment and expected `segy_major_revision`.~

No need to change since we seem to use `SEGY_BIN_SEGY_REVISION` and `SEGY_BIN_SEGY_REVISION_MINOR` instead of `SEGY_BIN_SEGY_REVISION_MAJOR` and `SEGY_BIN_SEGY_REVISION_MINOR`. 

I just leave the comment as reminder in case I need it later. 👴

applications/test/catb-dec.output
```diff
@@ -41,5 +41,5 @@ maxtrh    	32727
 survty    	32726
 timebc    	32725
 ntrst     	18446744073709551572
-ftroff    	18446744073709551571
+ftroff    	3600
```
ajaust, line 44: `ftroff` is changed so that segyio can read the file correctly (in tests?). The file has only a text header (3200 bytes) and a binary header (400 bytes). Therefore, the first trace is found at byte 3600. 

Is this correct?

lib/include/segyio/segy.h
```diff
@@ -285,7 +285,7 @@ typedef struct {
     int ext_textheader_count;
 
     /* Byte offset of the first trace in the file. */
-    long trace0;
+    unsigned long long trace0;
```
ajaust, line 288: I think we need to update the C++ experimental header as well. There are some other occurrences where it is `long long` instead of `unsigned long long` which might give some overflow issues when casting from unsigned to signed. I did a `rg trace0 | rg long` which find the following occurrences

```text
$ rg trace0 | rg long
mex/segyutil.c:    unsigned long long trace0;
mex/segyutil.h:    unsigned long long trace0;
python/segyio/segyio.cpp:    long trace0;
lib/src/segy.c:    long long pos = ds->metadata.trace0 +
lib/src/segy.c:    unsigned long long trace0;
lib/src/segy.c:    unsigned long long* trace0,
lib/src/segy.c:    long long trace0 = ds->metadata.trace0;
lib/experimental/segyio/segyio.hpp: * long trace0()     - offset of first trace past extended text headers
lib/experimental/segyio/segyio.hpp:    long trace0()     const noexcept(true);
lib/experimental/segyio/segyio.hpp:long trace_meta_fromfile< T >::trace0() const noexcept(true) {
lib/include/segyio/segy.h:    unsigned long long trace0;
lib/include/segyio/segy.h:    unsigned long long* trace0,
lib/test/segy.cpp:    long trace0 = 3600;
lib/test/segy.cpp:    const unsigned long long trace0 = fp->metadata.trace0;
lib/test/segy.cpp:        unsigned long long trace0 = fp->metadata.trace0;
lib/test/segy.cpp:    const long trace0     = 3600;
```

My interpretation:

- The occurences in `test/segy.cpp` are probably fine since it has the correct type or will be upcast from `long` to `unsigned long long` (which should be safe).
- Occurences in the C++ header should be updated.
- Receiving values from `trace0` in `lib/src/segy.c` might be safer as `unsigned long long`.
- Other occurences appear to have the correct type.

What do you think?

python/segyio/segy.py
```diff
@@ -307,20 +307,40 @@ def offsets(self):
...
+        """Number of extended text header blocks.
+
+        The number of extra text headers present in the file.
 
-        return self._ext_headers
+        Returns
+        -------
+        headers : int
+            Number of extra text headers
+        """
```
ajaust: I am considering if we should just use the phrasing from the SEGY standard. I did not check though if this is consistent with our phrasing in the rest of the documentation.
```text
        """Number of extended textual file header records.
        
        The number of 3200-byte, extended textual file header
        records following the binary header.
            
        Returns
        -------
        headers : int
            Number of extra text headers
        """
```
I just leave the comment here to think about it.

lib/src/segy.c
```diff
@@ -1675,153 +1675,97 @@ int segy_traceheaders(
...
-    };
-
-    const int bytes2_len = sizeof( bytes2 ) / sizeof( int );
-
-    for( int i = 0; i < bytes2_len; ++i ) {
-        const int offset = bytes2[ i ] - 1;
-        bswap16_mem( xs + offset, xs + offset );
-    }
 
+static int bswap_th(
```
ajaust, line 1727: I like how much simpler this function looks now. 🥳

lib/test/segy.cpp
```diff
@@ -1400,20 +1358,18 @@ SCENARIO( "modifying trace header", "[c.segy]" ) {
 namespace {
 
 int check_sorting( const std::string& name ) {
-    int trace0 = 3600;
```
yngve793, line 1403: I am a bit confused here. `trace0` and `trace_bsize` are reassigned in the old code but this does not seams to be replicated in the new code.

mex/segy_get_ps_line_mex.c
```diff
@@ -26,19 +26,23 @@ void mexFunction(int nlhs, mxArray *plhs[],
     int il = (int)mxGetScalar(mx_il_word);
     int xl = (int)mxGetScalar(mx_xl_word);
 
-    segy_file* fp = segyio_open( spec.filename, "rb" );
```
yngve793, line 29: Is this a bug fix or have we changed the name in this PR?

python/segyio/segyio.cpp
```diff
@@ -1059,9 +1053,7 @@ PyObject* getth( segyfd* self, PyObject *args ) {
                            SEGY_TRACE_HEADER_SIZE, buffer.len() );
 
     int err = segy_traceheader( ds, traceno,
```
yngve793: Can we have all parameters in one line?

ajaust (commented):
The changes look great. 🙂 

I left some remark w.r.t. changed data type of `trace0` which might be critical. Most other comments are mainly for my understanding and clarify our assumptions.

applications/segyinfo.c
```diff
@@ -61,7 +61,7 @@ int main(int argc, char* argv[]) {
     }
 
     char traceh[ SEGY_TRACE_HEADER_SIZE ];
-    err = segy_traceheader( fp, 0, traceh );
+    err = segy_read_standard_traceheader( fp, 0, traceh );
```
yngve793, line 64: I can see that it is useful to separate different trace headers with name.

lib/include/segyio/segy.h
```diff
@@ -632,6 +632,20 @@ int segy_read_ext_textheader( segy_datasource*, int pos, char* buf );
...
+int segy_read_traceheader( segy_datasource*,
+                           int traceno,
+                           int traceheaderno,
+                           const segy_entry_definition* mapping,
+                           char* buf );
+
+/* Write the 'traceheaderno' trace header at `traceno` from `buf` into file. */
+int segy_write_traceheader( segy_datasource*,
+                            int traceno,
+                            int traceheaderno,
```
yngve793: Consider: traceheaderno -> traceheader_no?

lib/src/segy.c
```diff
@@ -1675,153 +1675,97 @@ int segy_traceheaders(
...
+    if( ds->metadata.endianness != SEGY_LSB ) return SEGY_OK;
+
+    int offset = 0;
+    while( offset < SEGY_TRACE_HEADER_SIZE ) {
+        int datatype = segy_entry_type_to_datatype( mapping[offset].entry_type );
+        if( datatype == SEGY_UNDEFINED_FIELD ) {
+            ++offset;
+            continue;
+        }
+        int size = segy_formatsize( datatype );
```
yngve793, line 1741: The base structure is starting to pay off. :-)

yngve793 (approved):
It is nice to see that bits and pieces are starting to come together. 

This PR contains simpler and in general more basic changes. 
I have only added some minor comments.

lib/src/segy.c
achaikou: Khe-khem... :blush:   We already do what you show in the example  :innocent: 

See the last changed lines of this first commit. There we call
```
    return seek_traceheader_offset(
        ds,
        traceno,
        ds->metadata.traceheader_count,
        min * elemsize,
        trace0,
        trace_bsize
    );
```
So we do call it with `traceheader = traceheader_count` and seek inside of the samples.

I noticed it, didn't find a nice and quick way of dealing with it (what more can I rename this to ??) and decided that the caller knows what they are doing...

So it is not a bug, it is a feature :smile:

lib/test/segy.cpp
achaikou: New code does
`Err err = segy_collect_metadata( fp, -1, -1, 0 )`
which would collect these values directly from the file.
`trace0` is at standard position 3600 in all the files.
Also I see from `Readme` that all files use 10 float samples, so `10 * 4` and this information is also collected automatically now.

mex/segy_get_ps_line_mex.c
achaikou: This is a bug fix, but maybe I shouldn't fix that and revert it instead.

Basically `segyio_open` function doesn't exist. I have no clue why matlab code compiles.
This script is not used in any test, as far as I understand. I have no clue if this script was actually working.

So after your comment I decided that I probably should revert this line altogether. Lets this weird matlab thing stay weird. :woman_shrugging:

applications/test/catb-dec.output
achaikou: Correct!

lib/include/segyio/segy.h
achaikou: Oh, good. I thought I updated everything because no CI complained and they are usually pretty strict about these things.

Changing `long long` to `unsigned long long` in .c functions - probably no. 
- `seek_traceheader_offset` is a mix of many types already,
- `segy_traces` you anyway would require some casting, and current approach assures that if first bit is set (so it won't fit into long long), we will catch it comparing to 0.

So there I would prefer to keep long long.

Other missing stuff is updated :+1:

lib/include/segyio/segy.h
achaikou: Our current code that counts text headers `parse_extended_text_headers` is in C++ because we wanted nice C++ features.

Do you think we should duplicate the code to C?
I decided the easiest approach is to fail here.

I don't know what to do with this in the long run (switch everything to C core? drop the base of this project and drop C core completely?), but it is not the problem for our short-term horizon.

python/segyio/segy.py
achaikou: Probably makes sense :+1: 
Don't think we have any established pattern there.

python/segyio/segyio.cpp
achaikou: This functions gets changed in 2 next commits as well, resulting in 5 parameters again. So it is nicely formatted only in the last version.

Can fix all intermediate commits on rebase, if you want :cat:

achaikou (commented):
Thanks for the reviews! :smiley: 
Fixups are added.

I've already rebased changes to main and there were no controversial merge conflicts.

lib/test/segy.cpp
yngve793: ok

python/segyio/segyio.cpp
yngve793: No, need then.

lib/include/segyio/segy.h
ajaust: Thanks for the explanation. 🙂

ajaust (approved):
Looks good from my side. 👍

mex/segy_get_ps_line_mex.c
yngve793: OK, fine.

PR #652, opened 2025-10-28
Drop macos intel support

achaikou:
Macos Intel will end support soon. Github actions will drop the image.
As we prepare for 2.0 release, we can drop it already now and leave it be in the version-1.x branch only.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -56,10 +56,6 @@ jobs:
             build_type: "Debug"
             compiler_flags: "-Wextra -Wall -pedantic"
             build_python: "OFF"
-          - os: macos-13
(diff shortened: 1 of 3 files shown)


ajaust (approved):
Looks good.

Just a note in case we need it in the future:
By dropping the support for Intel Macs we also implicitly increase the minimum supported MacOS version to [Big Sur](https://en.wikipedia.org/wiki/MacOS_Big_Sur). This would allow us to bump the `-DCMAKE_OSX_DEPLOYMENT_TARGET` when building wheels (see `pyproject.toml`). This has some impact on what C++ standard we can use. I did not investigate what deployment target we could choose now and what this would imply w.r.t. C++ standard.

PR #653, opened 2025-10-28
Backport 1.x - initial backport

achaikou:
Changes gathering cut-off is commit 1e7a22d2135ab45fb9c80ff57f52b0aebead96fb

----

First PR against ~`backport-1.x`~ `segyio-1.x` branch. Branch name can be up for discussion, if we want something nicer here, as it is not just backport but 1.0 release. I already vetoed name `v1.x` as it is similar to tags and `release-1.x` as it is similar to one existing branch which might be used for something).
Having the separation done already now is sponsored by macos intel starting to break in November.

I don't know how backporting stuff is done properly and if it works the way I think it works, so lets figure it out.

PR Contains general commits from "main" branch. I tried to add there commits touching only workflows-tests-docs, probably not all of them anyway, but major ones. But no functional changes.

Publishing workflow is still a bit of a mystery, though looks like nothing specific would be required. We might need to test it on test-pypi before publishing 3.14 python release.

We probably don't need benchmarks workflow, but I picked commits with it for simplicity. Also it might make it a bit easier to switch base benchmarking reference to next segyio 1.0 release later?..

Btw, this PR truly shows the power of atomic commits :smile: 
I could cherry-pick commits from the last 6 months and get like two small merge conflicts :smiley: 
The further we would go, the worse problem would obviously become, but probably if that happens, `backport-1.x` branch should have only python version updates and nothing more.

See if you think I am missing some commits/added some commits that shouldn't be here/should have done this all differently.

### DIFF
--- a/.github/workflows/analyzers.yaml
+++ b/.github/workflows/analyzers.yaml
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/benchmarks.yaml
+++ b/.github/workflows/benchmarks.yaml
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
     inputs:
       commit_ref:
--- a/.github/workflows/bigendian.yaml
+++ b/.github/workflows/bigendian.yaml
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/large.yaml
+++ b/.github/workflows/large.yaml
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
 
 jobs:
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -3,11 +3,11 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [segyio-1.x]
     tags:
       - '*'
   pull_request:
-    branches: [main]
+    branches: [segyio-1.x]
   workflow_dispatch:
 
 jobs:


ajaust:
To me to the changes look good overall. I also don't know if there is any better way to keep a branch for segyio v1 than the way that you propose it here.

Also for understanding/alignment. I assume the workflow in the future for the `backport-v1.x` branch would be:

1. Create a new branch from `backport-1.x`.
2. Create commits, e.g., by cherry-picking, in the branch
3. Create PR targeting `backport-1.x`.
4. Merge PR into `backport-1.x`.

Is this correct?

# Remarks

## General

- I would consider adding a note to the README.md that segyio's v1 releases are maintained in a different branch than `main`.

## List of commits

I went through the commit history of `main` and the PR and found the following commits for where I wonder if these should be added.

### Before cutoff

#### October 16

d127d60b54077a98a93461eb44365703315c20ac Add additional tests to CI that build Python bindings

#### June 6

f041398c5341a4f88b9dae508b6a460d862da1ba Fix printf in segyio-catb


#### May 7

13f0be1a30139154604732047ba5844c0d722025   Add CPP check for macos 13
c0c9ef563d6cb9d2765f10217c3e62dad34c7adb Fix cppcheck exceptions on MAC

### After cutoff (too new)

I add these for completeness since I went through the full history. I am completely fine to add these later.

#### October 29

7bebfd7a486fe18ca955e20c57c4c8d551802df8 Reduce permissions of GitHub tokens in Actions

#### October 28

09fa7f33d284c8c757090facd92f45e3ea8aa725 Improve return value check

#### October 22

7b4470cb80801da06b332373db9cc501fe8473a3 Declare C++11 as used in whole project

README.md
```diff
@@ -357,10 +357,10 @@ discriminate between the revisions, but instead tries to use information
...
 publications by SEG:
 
-- [SEG-Y 0 (1975)](https://library.seg.org/pb-assets/technical-standards/seg_y_rev0-1686080980707.pdf)
-- [SEG-Y 1 (2002)](https://library.seg.org/pb-assets/technical-standards/seg_y_rev1-1686080991247.pdf)
-- [SEG-Y 2 (2017)](https://library.seg.org/pb-assets/technical-standards/seg_y_rev2_0-mar2017-1686080998003.pdf)
-- [SEG-Y 2.1 (2023)](https://library.seg.org/pb-assets/technical-standards/seg_y_rev2_1-oct2023-1701361639333.pdf)
+- [SEG-Y 0 (1975)](https://seg.org/wp-content/uploads/2025/11/seg_y_rev0.pdf)
+- [SEG-Y 1 (2002)](https://seg.org/wp-content/uploads/2025/11/seg_y_rev1.pdf)
+- [SEG-Y 2 (2017)](https://seg.org/wp-content/uploads/2025/11/seg_y_rev2_0_mar2017.pdf)
+- [SEG-Y 2.1 (2023)](https://seg.org/wp-content/uploads/2025/11/seg_y_rev2_1-oct2023.pdf)
```
ajaust, line 363: Should we also link to the [SEG-Y_r2.1: Extended Textual Header Clarification"](https://seg.org/wp-content/uploads/2025/11/seg-y_rev2-1_clarification_2025-05-01.pdf)?

.github/workflows/analyzers.yaml
```diff
@@ -3,9 +3,9 @@ permissions: {}
 
 on:
   push:
-    branches: [main]
+    branches: [backport-1.x]
```
ajaust: I am unsure if it would have any advantage (or disadvantage) to leave the `main` branch as potential target to trigger the workflow.
```suggestion
    branches: [main, backport-1.x]
```

ajaust (approved):
Looks good overall. I just added some two small comments to the commits.

Below, replies to your comment. 🙂 
---

Workflow :+1:

---

Picked commits 👍 

---

Note to README:
Yes, I meant that we should add a note to the `README.md` in the `main` branch. The intention is to make it visible to users where the v1 code is located. 

Changelog:
I think we should have an updated changelog in both branches since we plan to have a release for both branches as well. This makes it simple for users to get a quick overview of the changes changes depending on the release.

I agree, making a release/adding a changelog is another PR.

---

Yes, I like `segyio-1.x` as branch name. Just update the GitHub Actions accordingly

---

I don't know what type of references are generated by the cherry-pick. I have only checked the commits on GitHub and there does not seem any reference to your fork or I did not see it.

.github/workflows/analyzers.yaml
achaikou: I think it would only become more confusing.

We do not intend to keep workflows between branches 100% in sync. So we want the most complete version of the workflow to run on the "main" operations and somewhat reasonable to run on "backport" operations.

To me that means that each branch should have its own workflow file with only desired branch name.

I am also not 100% how github would handle having "main" workflow here. I don't think it would run both workflows as this commit will not belong to main branch, but it all would just confuse me. :smile:

README.md
achaikou: We merged the original commit without it, so probably no.

Also here are the links to the specifications only, so probably no need to add clarification too.

ajaust:
I marked the last open comments as resolved. 🙂

PR #654, opened 2025-10-30
Use trusted publisher for PyPI publication

ajaust:
The branch contains several commits (currently the last three) to have a simpler workflow for testing publication to TestPyPI. This breaks some of the windows tests. 

Successful publication to TestPyPI happened in [this GitHub Action workflow](https://github.com/ajaust/segyio/actions/runs/18943837389). The pushed package is [in this TestPyPI repository](https://test.pypi.org/project/test-segyio-2025/). Relevant documentation about [publishing using trusted publishers](https://docs.pypi.org/trusted-publishers/using-a-publisher/).

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -70,6 +70,13 @@ jobs:
     name: Publish wheels to PyPI
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+
+    environment:
+      name: pypi
+
+    permissions:
+      id-token: write
+
     steps:
       - uses: actions/checkout@v4
 
@@ -80,9 +87,6 @@ jobs:
           path: ./wheelhouse/
 
       - name: Publish wheels to PyPI
-
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          user: __token__
-          password: ${{ secrets.PYPI_API_TOKEN }}
           packages-dir: ./wheelhouse/


.github/workflows/wheels.yaml
```diff
@@ -90,3 +90,31 @@ jobs:
...
+        with:
+          pattern: wheel-*
+          merge-multiple: true
+          path: ./wheelhouse/
+
+      - name: Publish wheels to PyPI
+        uses: pypa/gh-action-pypi-publish@release/v1
+        with:
+          packages-dir: ./wheelhouse/
+          repository-url: https://test.pypi.org/legacy/
```
ajaust: This should be almost a 1 to 1 copy of our intended workflow for publishing to PyPI to verify that the workflow makes sense. Only things needed to publish to TestPyPI were changed.

.github/workflows/wheels.yaml
```diff
@@ -70,6 +70,13 @@ jobs:
     name: Publish wheels to PyPI
     runs-on: ubuntu-latest
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+
+    environment:
+      name: pypi
```
achaikou, line 75: To confirm:
- Our repository doesn't even have `pypi` environment.
- Nothing seems to use it and yet because in testpypi you define a different one, something must require it.
- This something must be `pypi` itself where you link via environments. Or not? The other [link](https://docs.pypi.org/trusted-publishers/using-a-publisher/) states "Specifying a GitHub environment is optional, but strongly encouraged". But I assume toolbox specified the environment there, so we must use it too.
- we need url for test-pypi but don't need it for pypi? It seems present in the [example](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#the-whole-ci-cd-workflow).
- Docs say something about `For security reasons, you must require manual approval on each run for the pypi environment.` Would we need to do something??

?

ajaust: I understand your doubts. 🙂 I hope this can clarify open questions:

### GitHub environment

- Yes, we are lacking the environment on GitHub. I will add one, but I was hesitant to add it before we are happy with the content of this PR. I added it now with the default options (=no additional restrictions).
- Yes, in general, an environment is not required. However, if specified, PyPI can do some check and reject publications pushed from the from the wrong or no environment. 
- An environment allows us to add some additional safeguards on the GitHub side (delay the workflow, require specific reviewers, specify secrets only visible inside the environment). I don't think we need any of the additional safeguards from the GitHub side at the moment.
- The toolbox was open to not specify an environment, but I asked them to specify the `pypi` environment in the config. It is the current best practice and allows us to add more restrictions later on if we need them without the need to ask the toolbox to update anything on their side.
- I attach a screenshot at the bottom that shows the TestPyPI configuration panel. It looks the same on PyPI. This gives an idea what information PyPI verifies when using. trusted publishers.
- Regarding potential restrictions defined by environments on the GitHub side you check out the `pypi` envornment that I just added. Maybe you find an option that is useful to configure.

### URL

- The "normal" PyPI is the default URL of the GitHub Action. Therefore, we do not need to specify it. 
- The example you link sets `url` in the environment, but not the `repository-url` of the `pypa/gh-action-pypi-publish` for the publication to PyPI. The `repository-url` is the relevant target for uploading the wheels.
- The example also states that the `url` is specified to nicer output in the GitHub GUI (?)
  > This also defines a GitHub Environment for the job to run in its context and a URL to be displayed in GitHub’s UI nicely.
  [Source](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#defining-a-workflow-job-environment)

- I don't remember how the output looked for `si4ti` and the last workflow run is too old that the logs are deleted.
-  [This documentation](https://docs.pypi.org/trusted-publishers/using-a-publisher/) does not mention the `url` part of the environment at all.

### Manual approval

- I think the phrasing is off in the docs. We can configure manual approvals of the workflow as part of the environment configuration, but we do not have to. [This blog post](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/#additional-security-hardening-is-available) has the correct formulation in my opinion.
- I did not configure manual approval of the workflow in `si4ti`.
- There is [discussion of security related options](https://docs.pypi.org/trusted-publishers/security-model/) to additionally secure the publication flow, e.g., via manually approved workflow runs. From my understanding, most security concerns exist for repositories with many users in several roles where one wants to ensure that the repository owners/admins are somehow in the loop of a publication flow.
- I am not against enabling manual approvals or other restrictions, I just did not explore them too much. I enabled "manual approvals" now and mapped it to the `one::pac` group. We could give it a try or are you worried that this might be to disruptive?

### Working workflow on si4ti

Just for reference, this [workflow successfully](https://github.com/equinor/si4ti/blob/343ad15f7ea82e88c5af4a6995e05e68c2dbf3d0/.github/workflows/publish.yml#L167-L188) publishes `si4ti` packages.

### TestPyPI publisher configuration

Screenshot 2025-11-12 at 15 03 42

PR #655, opened 2025-11-06
Rev2: types and functions

achaikou:
- get/set field and field_forall are supported
- floats and strings (for header names) are supported
- all data is available in python, but python's user interface is not updated yet.


python/segyio/segyio.cpp
```diff
@@ -1662,6 +1662,19 @@ PyObject* getdt( segyfd* self, PyObject* args ) {
     return Error( err );
 }
 
+PyObject* getdelay( segyfd* self ) {
```
ajaust, line 1997: In the C interface it is called `segy_delay_recoding_time`. I am wondering if we should align the names to `getdelayrecordingtime` (or similar) here, `segy_delay` in C or use some name in between. 

I must admit that I don't have the full overview how consistent names are between Python and C. I see that the consistency in the Python interface is not guaranteed as some methods, e.g., `rotation`, are not prefixed with `get`.

lib/src/segy.c
```diff
@@ -1846,6 +1846,84 @@ int segy_traces( segy_datasource* ds,
     return SEGY_OK;
 }
 
+/* Allows only int2 and uint2. */
+static int field_data_to_scalar(
```
ajaust: Do we expect this method to become more generalised? If not, should the restriction (only be usable for 2 byte ints) be reflected in the method's name?

lib/src/segy.c
```diff
@@ -1846,6 +1846,84 @@ int segy_traces( segy_datasource* ds,
...
+    switch( delay_fd.entry_type ) {
+        case SEGY_ENTRY_TYPE_TIME2:
+        case SEGY_ENTRY_TYPE_INT2:
+            delay_raw = delay_fd.value.i16;
+            break;
+        case SEGY_ENTRY_TYPE_UINT2:
+            delay_raw = delay_fd.value.u16;
+            break;
+        default:
+            return SEGY_INVALID_FIELD_DATATYPE;
```
ajaust, line 1986: Does it make sense to allow for `SEGY_ENTRY_TYPE_TIME2` such that we could call `field_data_to_scalar` here?

I am mainly wondering if we would have a good way to reuse `field_data_to_scalar` here instead of reimplementing (almost) the same method here.

lib/src/segy.c
```diff
@@ -1846,6 +1846,84 @@ int segy_traces( segy_datasource* ds,
...
+    if( err != SEGY_OK ) return err;
+
+    float delay_scalar;
+    if( scalar_raw == 0 ) {
+        delay_scalar = 1.0f;
+    } else if( scalar_raw < 0 ) {
+        delay_scalar = -1.0f / scalar_raw;
+    } else {
+        delay_scalar = 1.0f * scalar_raw;
+    }
```
ajaust: We probably don't need it here yet, but maybe this could become its own method. I assume that is the reconstruction of a `time2` value. In this sense, it could be reused other places later.

python/segyio/open.py
```diff
@@ -275,13 +275,11 @@ def _open(datasource_descriptor,
...
-        if delay_scalar == 0:
-            delay_scalar = 1
-        elif delay_scalar < 0:
-            delay_scalar = 1.0 / delay_scalar
         dt = segyio.tools.dt(f, fallback_dt = 4000.0) / 1000.0
-        t0 = f.header[0][segyio.TraceField.DelayRecordingTime] * abs(delay_scalar)
+        try:
+            t0 = fd.getdelay()
+        except RuntimeError:
+            t0 = 0.0
```
ajaust, line 282: Doesn't the change modify the behavior by catching the runtime error and setting `t0` to `0` compared to before?

lib/src/segy.c
```diff
@@ -1846,6 +1846,84 @@ int segy_traces( segy_datasource* ds,
...
+    const int delay_offset = standard_name_map[SEGY_TR_DELAY_REC_TIME];
+    int delay_raw = 0;
+    segy_field_data delay_fd;
+    err = segy_get_tracefield(
+        trace_header, standard_map, delay_offset, &delay_fd
+    );
+    if( err != SEGY_OK ) return err;
+
+    // until someone complains, we restrict allowed types to these only
+    switch( delay_fd.entry_type ) {
```
yngve793, line 1977: This can be combined into field_data_to_scalar.

python/segyio/segyio.cpp
```diff
@@ -1135,6 +1135,229 @@ PyObject* putth( segyfd* self, PyObject* args ) {
     }
 }
 
+PyObject* getfield( segyfd* self, PyObject* args ) {
```
yngve793, line 1146: The diff is not very big. Is it possible to keep it in the same place? In this way git will be able to detect the difference. Alternatively it can be moved afterwards in a separate commit.

python/segyio/field.py
```diff
@@ -374,10 +374,12 @@ def __getitem__(self, key):
         { 37: 5, 189: 2484 }
         """
 
-        try: return self.getfield(self.buf, int(key))
+        traceceheader_no = 0
```
yngve793: Consider changing to traceceheader_nr? I presume that the postfix indicate number?
My initial thought is that this is Norwegian[no]  traceceheader.
I get the feeling that a native Norwegian speaker has created this name.

lib/src/segy.c
```diff
@@ -1508,28 +1503,9 @@ static int bswap_bin( const segy_datasource* ds, char* xs ) {
...
-            case 2:
-                bswap16_mem( xs + offset, xs + offset );
-                break;
-            case 1:
-                break;
-            default:
-                return SEGY_INVALID_FIELD_DATATYPE;
-        }
-        offset += size;
+        int next = bswap_header_field_value( binmap, xs, offset );
```
yngve793, line 1588: Very nice.

lib/src/segy.c
```diff
@@ -1734,28 +1710,9 @@ static int bswap_th(
...
-            case 2:
-                bswap16_mem( xs + offset, xs + offset );
-                break;
-            case 1:
-                break;
-            default:
-                return SEGY_INVALID_FIELD_DATATYPE;
-        }
-        offset += size;
+        int next = bswap_header_field_value( mapping, xs, offset );
```
yngve793, line 1795: :-)

lib/src/segy.c
```diff
@@ -1210,6 +1232,7 @@ static int set_field( char* header,
             memcpy( header + offset, &( fv.i8 ), vsize );
             return SEGY_OK;
 
+
```
yngve793: Unnecessary newline.

lib/src/segy.c
```diff
@@ -1097,6 +1097,7 @@ static int get_field( const char* header,
             memcpy( &(fd->value.i8), header + offset, vsize );
             return SEGY_OK;
 
+
```
yngve793: Unnecessary newline.

lib/src/segy.c
```diff
@@ -1388,42 +1435,71 @@ static int slicelength( int start, int stop, int step ) {
...
-
-        case SEGY_SIGNED_INTEGER_4_BYTE:
-            fd->value.i32 = bswap32( fd->value.i32 );
-            return SEGY_OK;
+/* Byte-swaps field value in the header. Returns new offset after swapping or -1
+ * if error occurred.
+ */
+static int bswap_header_field_value(
+    const segy_entry_definition* mapping,
+    char* xs,
```
yngve793, line 1439: What is `xs` short for?

lib/src/segy.c
```diff
@@ -1477,6 +1546,7 @@ int segy_field_forall( segy_datasource* ds,
                 memcpy( buf, &fd.value.i8, elemsize );
                 break;
 
+
```
yngve793: New line?

python/segyio/field.py
ajaust: Maybe we should consider defining an enum or so to give the `0` a meaning. I guess `0` would always refer to the standard trace header. Not sure about a good name, but I guess reading something like

```python
try: return self.getfield(self.buf, SEGY_TRACE_HEADER_INDEX.STANDARD, int(key))
```

or

```python
try: return self.getfield(self.buf, STANDARD_TRACE_HEADER, int(key))
```

would explicitly state which trace header we access here.

python/segyio/segyio.cpp
```diff
@@ -1218,7 +1239,7 @@ PyObject* putfield( segyfd* self, PyObject *args ) {
             if( offset < 0 || offset >= SEGY_BINARY_HEADER_SIZE ) {
                 return KeyError( "Invalid field %d", field );
             }
-            const segy_entry_definition* map = segy_binheader_map();
+            map = segy_binheader_map();
```
ajaust, line 1262: Note to self: Technically we do not need this since when data is set, we use `segy_set_binfield`. `segy_set_binfield` knows which mapping to use so it does not take `map` as input parameter. However, I prefer that we set map to the binary header map. It ensures that `map` will point to a valid header mapping when reaching line 1264.

python/test/segyio_c.py
```diff
@@ -164,13 +164,13 @@ def test_read_binary_header_fields(mmap=False):
     binary_header = f.getbin()
 
     with pytest.raises(TypeError):
-        _ = f.getfield([], 0)
+        _ = f.getfield([], 0, 0)
```
ajaust: I have to think a bit about adding `, 0` at so many places. I am a bit worried that we might forget what `0` means here at some point. Maybe we should assign it to a variable with reasonable name or define it as a constant within the test `STANDARD_TRACE_HEADER_IDX = 0` to use it. 
```python
_ = f.getfield([], STANDARD_TRACE_HEADER_IDX, 0)
```

However, I don't want overly long names either.

lib/src/segy.c
```diff
@@ -1424,6 +1424,8 @@ static int bswap_header_word( segy_field_data* fd ) {
 }
 
 int segy_field_forall( segy_datasource* ds,
+                       int traceheader_index,
+                       const segy_entry_definition* offset_map,
```
ajaust, line 1498: Do we need to pass both the index and the map? The method has access to the data source already so it could obtain the specific offset map from `ds` and the `traceheader_index`. What am I missing/misunderstanding?

python/test/segyio_c.py
```diff
@@ -443,30 +443,69 @@ def mkempty():
...
-    extension1_th = f.getth(0, 1, mkempty())
-    assert f.getfield(extension1_th, 0, field) == 42
+    with pytest.raises(KeyError):
+        f.getfield(mkempty(), nonexistent_index, field)
+
+    with pytest.raises(KeyError):
+        f.field_forall(attrs, nonexistent_index, start, stop, step, field)
+
+    with pytest.raises(KeyError):
+        f.field_foreach(attrs, nonexistent_index, indices, field)
```
ajaust, line 498: I am wondering if it would make sense to move this part of into its own test. This parts seems to be independent from the `verify_traceheader` part and also relies on variables (`start, `stop`...) which are exclusively used for reading from an nonexistent header.

python/segyio/segyio.cpp
```diff
@@ -2715,6 +2718,22 @@ int parse_py_TraceHeaderLayoutEntry_list( PyObject* entries, segy_header_mapping
...
+        int err = set_mapping_offset_to_entry_defintion(
+            mapping, "header_name", header_name_offset, "string8", false
+        );
+        if( err != SEGY_OK ) return err;
+    } else {
+        // we accept that users could have written their own values here only for SEG00000
+        if( strncmp( mapping->name, "SEG00000", 8 ) != 0 ) {
+            return SEGY_INVALID_ARGS;
+        }
+    }
```
ajaust, line 2812: I don't understand yet what we are doing here nor why.

After the mapping has been extracted, we check the mapping at the `header_name_offset` location. 

If the entry type is `SEGY_ENTRY_TYPE_UNDEFINED`, this indicates that the header name is stored at this location and we update the mapping accordingly.

If the entry type is **not** `SEGY_ENTRY_TYPE_UNDEFINED`, we check if the name of the field at the `header_name_offset` is `SEG00000`. If the name is `SEG00000`, we accept it, but other names are rejected. 

Is anything I write correct? If it is correct, I think I am still not sure why we need to do this check.

lib/src/segy.c
```diff
@@ -1389,39 +1389,35 @@ static int slicelength( int start, int stop, int step ) {
...
-            fd->value.u16 = bswap16( fd->value.u16 );
-            return SEGY_OK;
-
-        case SEGY_UNSIGNED_CHAR_1_BYTE:
-        case SEGY_SIGNED_CHAR_1_BYTE:
-            return SEGY_OK;
+/* Byte-swaps field value in the header. Returns new offset after swapping or -1
+ * if error occurred.
+ */
+static int bswap_header_field_value(
```
ajaust, line 1437: Nice! I like this. :)

lib/src/segy.c
```diff
@@ -1117,11 +1117,34 @@ static int get_field( const char* header,
...
+            val = be32toh( val );
+            memcpy( &( fd->value.f32 ), &val, vsize );
+            return SEGY_OK;
+        }
+
+        case SEGY_IBM_FLOAT_4_BYTE: {
+            uint32_t val;
+            memcpy( &val, header + offset, vsize );
+            val = be32toh( val );
+            ibm_native( &val );
```
ajaust, line 1139: Could you confirm that the order of operations is correct?

1. Swap bigendian to host endianness
2. Convert from IBM float to native float

I don't mean to question the code is correct, but I need some reassurance. I looked into the `ibm_native` code I don't understand what is going on so I am scared now. 😨

lib/src/segy.c
```diff
@@ -1117,11 +1117,34 @@ static int get_field( const char* header,
             memcpy( &(fd->value.u8), header + offset, vsize );
             return SEGY_OK;
 
-        case SEGY_IEEE_FLOAT_8_BYTE:
-            memcpy( &(fd->value.f64), header + offset, vsize );
-            memcpy( &val, &(fd->value.f64), sizeof( uint64_t ) );
```
ajaust, line 1121: So we just had some unnecessary copy here? Nice catch. :)

python/segyio/segyio.cpp
```diff
@@ -2668,6 +2688,41 @@ int set_mapping_offset_to_entry_defintion(
...
+        char* mant_name_heap = new char[mant_name.size() + 1];
+        if( !mant_name_heap ) return SEGY_MEMORY_ERROR;
+        std::strcpy( mant_name_heap, mant_name.c_str() );
+
+        std::string exp_name = spec_entry_name + "_exp";
+        char* exp_name_heap = new char[exp_name.size() + 1];
+        if( !exp_name_heap ) return SEGY_MEMORY_ERROR;
+        std::strcpy( exp_name_heap, exp_name.c_str() );
+
+        delete[] spec_entry_name_heap;
```
ajaust: Just a though. If we are getting worried about pointers and memory leaks, we could use smart pointers. This might worth considering if we see that explicit heap memory allocations become more common.

Example (did not test):

```c++
std::unique_ptr<char> spec_entry_name_heap(new char[spec_entry_name.size() + 1]);
// Code...
if (entry_type == SEGY_ENTRY_TYPE_SCALE6_MANT) {
  // Code ...
  // No explicit deletion needed as unique_ptr goes out of scope and deletes the object it owns
  return SEGY_OK;
}

segy_entry_definition def = {
  entry_type,
  requires_nonzero_value,
  spec_entry_name_heap.release() // Releases ownership and returns pointer to managed object
};
mapping->offset_to_entry_definition[byte - 1] = def;
return SEGY_OK;	    return SEGY_OK;
```

This can also be applied to objects on heap that we plan to hand over to Python.

lib/test/segy.cpp
```diff
@@ -1298,6 +1298,84 @@ TEST_CASE( "setting correct header fields succeeds",
...
+
+TEST_CASE( "setting correct header fields of string type", "[c.segy]" ) {
+    char header[SEGY_TRACE_HEADER_SIZE] = {};
+    std::string value = "TEST";
+
+    SEGY_ENTRY_TYPE entry_type = SEGY_ENTRY_TYPE_STRING8;
+
+    segy_field_data fd;
+    fd.entry_type = entry_type;
+    memset( fd.value.str8, 0, 8 );
```
ajaust: I am not sure if we have any preference or clear style for strings. We could set the values to `\0` explicitly to point out that this is a string. 

```c++
 memset( fd.value.str8, '\0', 8 );
```

test-data/README.md
```diff
@@ -105,4 +107,6 @@ Files are created manually.
 | mapping-multiple-stanzas.sgy   | Other stanzas are present before and after layout stanza.     |
 | mapping-shifted.sgy            | Default mapping with all offsets shifted by 1 byte.           |
 | mapping-unparsable.sgy         | XML is unparsable.                                            |
+| mapping-unparsable-lsb.sgy     | XML is not ascii. Not related to 'mapping-unparsable.sgy'     |
```
ajaust: How is the filename `mapping-unparsable-lsb.sgy`, which points me to least significant bit (lsb) , related to the description which says `XML is not ascii`, which I interpret as alternative encoding (EBCDIC, UTF or something else). If the file is using the least significant bit byte order, I would assume that valid ASCII is still defined (after byte swapping).

python/examples/flip-endianness.py
```diff
@@ -0,0 +1,52 @@
+import sys
+import segyio
+
+
+def main():
```
ajaust, line 5: Why do we need this if we have `flip-endianness` already in `applications/`?

test-data/README.md
```diff
@@ -96,6 +96,8 @@ Files are created manually.
 
 | File                           | Purpose                                                       |
 |--------------------------------|---------------------------------------------------------------|
+| mapping-all-types.sgy          | Entries are given all types possible.                         |
+| mapping-all-types-lsb.sgy      | `flip-endianness.py mapping-all-types.sgy out.sgy`            |
```
ajaust, line 100: Why not use the `flip-endianness` application in `applications/`?

python/test/segyio_c.py
```diff
@@ -443,30 +443,69 @@ def mkempty():
...
+    standard_index = 0
+    extension1_index = 1
+    proprietary_index = 2
+    nonexistent_index = 3
+
+    expected = {
+        standard_index: [0x11111111, 0x44444444],
+        extension1_index: [0x22222222_22222222, 0x55555555_55555555],
+        proprietary_index: [0x3333, 0x6666],
+    }
```
ajaust, line 486: We could consider making the [test parametrized](https://docs.pytest.org/en/7.1.x/example/parametrize.html) and having the `header_index` and the expected output as paremeters.

python/test/segy.py
```diff
@@ -2274,3 +2280,133 @@ def test_trace_header_extensions():
...
+    def test(f):
+        def mkempty():
+            return bytearray(240)
+
+        types_layout = f._traceheader_layouts["TYPES"]
+        for trace_index, header_cases in types_headers.items():
+            for case in header_cases:
+                offset = types_layout.entry_by_name(case.name).byte
+
+                # assert f.traceheader[trace_index][traceheader_index][offset] == case.opened
```
ajaust, line 2368: Are the assertions that are commented out going to be used later when the Python interface is in place?

python/test/segy.py
```diff
@@ -2274,3 +2280,133 @@ def test_trace_header_extensions():
...
+        # scale 6 type is inconsistent and not supported well yet
+        HeaderTest("scale6_mant", -1018, 1018),
+        HeaderTest("scale6_exp", -1019, 1019),
+        HeaderTest("header_name", b"TYPES\x00\x00\x00", b"TYPES   "),
+    ]
+
+    types_headers = {
+        0: types_header0_cases,
+        1: types_header1_cases,
+    }
```
ajaust: This refers to the (extended?) trace header (=`tracehader_index = 1`) for two traces (`trace[0]` and `trace[1]`). Is this correct?

From the naming, I was initially expecting that we would read the standard trace header (index `0`) and the first extended trace header (index `1`).

python/examples/flip-endianness.py
```diff
@@ -0,0 +1,52 @@
...
+
+        with segyio.create(dstfile, spec) as dst:
+            for i in range(1 + src.ext_headers):
+                dst.text[i] = src.text[i]
+
+            dst.bin = src.bin
+            dst.trace = src.trace
+
+            # intention:
+            # dst.traceheader = src.traceheader
```
yngve793, line 36: Can this line be removed?

python/segyio/segyio.cpp
```diff
@@ -1132,32 +1142,318 @@ PyObject* putth( segyfd* self, PyObject* args ) {
...
     int start, stop, step;
+    uint16_t traceheader_index;
     int field;
 
-    if( !PyArg_ParseTuple( args, "Oiiii", &bufferobj,
-                                          &start,
-                                          &stop,
-                                          &step,
-                                          &field ) )
+    if( !PyArg_ParseTuple(
```
yngve793, line 1437: More logical indent. 👍

ajaust (commented):
I think the changes look good overall. I added some remarks about naming. Just have a look how you feel about it. Otherwise, I added some questions where I got lost and could not figure out what is going on. 🙂

lib/src/segy.c
achaikou: No, I think this method is generalized enough.
No, I don't think restriction should be in the method's name. There are 4 scalars present in the specification, all 2 bytes.
PO decided that we should accept scalars only of this size.
Changing the name would imply other options are possible.

If someone complains in the future - then we shall look at it again.

lib/src/segy.c
achaikou: I don't understand.

We allow `SEGY_ENTRY_TYPE_TIME2` as the first case in this switch.

And I am not sure what you are wondering about.
PO wanted to allow here time2/int2/uint2 and treat those values as if they always were time2.

I'll add a comment.

yngve793 (commented):
Setting it to `Comment` for now. I have no critical comments, but it would be good to leave it for a couple of days before having a final look.

lib/src/segy.c
achaikou: 1. But delay is not a scalar (there are just 4 scalars in the specification), so this would be a lie.
2. Delay can (and in theory should be) time2. Scalar is not allowed to be time2.

lib/src/segy.c
achaikou: It is used in one other place, so I extracted the method :+1:

python/segyio/open.py
achaikou: I don't think it does in any meaningful sense.
`f.header[0]` gets first trace - if file is bad and we can't read the trace we would likely have seen errors earlier during metrics calculations (or definitely in `dt`).
`f.header[0][value]` - this may fail only if something is wrong with mapping. And I think in segyio 1.0 it would not fail but just return 0.

Anyway I did this because "delay" thing is just a property to nicely assemble `samples` array.
It probably is a bad idea to fail "open" because delay cannot be read correctly due to some type mismatch...

python/segyio/segyio.cpp
achaikou: There is no consistency or pattern whatsoever.
I called this `getdelay` because neighbor function I based this one upon (and the one used in the same `open` calculation) is called `getdt` and main operation in it is called `segy_sample_interval`.

`getdelayrecordingtime` or `get_delay_recording_time`  are too long compared with other methods on `segyfd`.
`dt` and `delay` are both official names of entries. So at least naming is consistent between those two :smile: 

Don't think it makes sense to unify stuff now more.

python/segyio/segyio.cpp
achaikou: I didn't want to technically split this commit into two and tried to do only minimal-required changes during move. But I agree, it is difficult to review anyway :(

Option 1:
I would prefer to move the code to proper place eventually anyway, I can split this commit into two technical ones.

1. Forward-declare two functions in their new place. Re-declare fd namespace in the old position. Do all other changes.
2. Remove forward-declarations and fd-namespace redeclaration from previous commit  and technically perform function move the to new position.

Option 2:
Alternative is to run this in the command line
```
git diff --color-moved 0116f974dfaec365ec735d15350be3fd3e25e030 8d60d1029d6d3c418cd17d1086f6857c7c09384e
```
This would color lines that were moved. Makes it relatively easy to not focus on just moved code.

Doing this I can see now that I added more than I intended, probably copy from wrong place.
I removed unintended code in a fixup, so thanks!! :heart: 

If you want, I can also use Option 1 and split into two commits on rebase.

python/segyio/field.py
achaikou: :norway: :partying_face: 

I was using it because of already used `traceno` in some places.
`_nr` I think is not used anywhere, but as in other places I used `traceheader_index` I will switch to that one (and fix the trace**ce**header typo :woman_facepalming: )

achaikou: 0 is the only traceheader that we would refer to by name, so it won't be an enum, but a single named constant.
So I don't think it is better than named variable :woman_shrugging: 

Also what makes sense in python code can be determined only after we figure out the best design for new python features and see if those functions would be reused or not...
Maybe we can discuss again then, with more knowledge, if it still bothers us?

python/test/segyio_c.py
achaikou: Makes sense.
Created a constant `SEG00000_INDEX` - short enough.
If you are not happy with this, will revert or change!

lib/src/segy.c
achaikou: The same thing I missed during demo run of this solution in the summer.
`ds` doesn't have all the maps. The maps are only in c++ binding code as we read them from xml as a vector of unknown length...

After I noticed this like 2 weeks ago I was wondering if we could copy the data to `ds` too. I think I decided against it in the end. Probably because this would be needed only in `read/write trace` and `segy_field_forall` functions.
So it will not help us much here, but might make flow between C++ and C even more complex.

python/test/segyio_c.py
achaikou: Not sure.
Each tests will either have to repeat non-existing-index tests or they will have to be a separate test "test_something_nonexisting" ?

To me all that looks like it tests one functionality, so I would leave it as is.
But let me know if you prefer it done differently and see a nice structure.

python/test/segyio_c.py
achaikou: It is dependent, there is `field_forall` in `verify_traceheader` too.

python/segyio/segyio.cpp
achaikou: You understand it correctly. :+1: 
I hoped I explained it good enough in the commit description, but let me put an extensive comment here too then.

lib/src/segy.c
achaikou: I never really figured that out :smile: 

`xs` is used very often in both segyio and dlisio.
My best guess it means `x` = unknown value `s` = many unknown values.
Copilot seems to agree

> It typically stands for "a collection of x" (where x is some element), with the s denoting plural.

I used it here as I think I copied `bswap_th` function that also uses `xs`.

lib/src/segy.c
achaikou: I added them on purpose to visually separate signed int - unsigned int - all other stuff, but will remove :+1:  Anyway formatter complains too...

lib/src/segy.c
achaikou: :hugs: 
I also have no clue what is happening there and have no wish to know :smile: 

I based my logic on tracing the calls of one place where `ibm_native`/`native_ibm` is used.

Logic leading to use of `ibm_native` seems to be:
- read bytes
- if file is little-endian, byteswap so that data is always big-endian
- if host is little endian, byteswap again, so that data endianness matched host-endianness
- turn ibm to nice float

and for `native_ibm`
- turn nice float to ibm
- if host is little endian, byteswap, so that data endianness is big-endian
- if file is little endian, byteswap again, so that data endianness matched expected file-endianness
- write bytes

New code follows the same logic:

In `get_field`:
- we assume that data given to us is big-endian, which is property of the library
- if host is little endian, byteswap, so that data endianness matched host-endianness
- so next step must be turning ibm to nice float.

In `set_field` then:
- we should turn nice float to ibm
- if host is little endian, byteswap, so that data endianness is big-endian
- returned field value is assured to be big-endian.

Honestly I am not 100% sure logic in `ibm_native`/`native_ibm` is correct. segy specification defines ibm layout in big-endian, but I just assume that conversion code assumes ibm and ieee float are of same endianness and does some magic.

But I decided that I don't have time now to dig for information how to convert the values and break my brain verifying conversion correctness. 
I assume that if it worked for many years, this logic must be correct. Or at least we would be consistently incorrect and this is problem for later. :smile: 

(but this was also part of the reason why I added little-endian test file and we have big-endian-host test workflow. I hope they would tell me if we screwed up...)

lib/test/segy.cpp
achaikou: No preference, no style. Done :+1:

python/segyio/segyio.cpp
achaikou: Not fully sure what you meant by "objects we plan to hand over to Python", assumed this is about mantissa and exponent parts. 

Something updated. You are right that we should try to use C++ when in .cpp. :+1:

python/examples/flip-endianness.py
achaikou: Because `flip-endianness.cpp` does not support mappings. You would need to give the application mappings data somehow and I have no clue how to (parsing xml in cpp seems difficult).
Also we agreed all applications are kind-of revision-1 support only.

Implementing the same via python was way easier.
As it is closer to the user, I didn't want to reference some already existing .c application which python users  have no clue about.

python/examples/flip-endianness.py
achaikou: I left it as "intended" line. I hope I would partly use it instead of current code once we implement python interface (hopefully next PR)

python/test/segy.py
achaikou: That is correct. Maybe I should remove word "types" (which was for all-types). Maybe it is confusing.

python/test/segy.py
achaikou: Yes.
We are not supposed to be using internals, those are difficult to read. But we don't have python interface in place yet, so I just wanted to be helpful and show what internals stand for :)

test-data/README.md
achaikou: Can't be used for files with custom mappings.

test-data/README.md
achaikou: Struggled a lot with this description and still missed :smile: 

I was wondering to add this file or not add it at all.
It was a file that I accidentally created and which also accidentally somewhere along development caught one bug related to two things at the same time:
1. File being little-endian and thus number of extended text headers being written in little endian, support for which was not implemented properly then, so value was read as 255 or something, and
2. XML layout not being written in ASCII and thus being unparsable trash (as we assume layout stanza is always ASCII).

Something clashed somewhere and instead of failing nicely it caused a nasty segfault.
So I was very uncertain if I should just delete it or not... Deciding that unlikely test is better than no test, I added it in the end, but struggled a lot with explanation why I did it :smile: 

If you have a nice short description that describes situation better, please suggest. I was fighting this whole thing for an hour already :smile:

achaikou (commented):
Thanks for the fast reviews!! :heart_eyes_cat:

python/segyio/segyio.cpp
yngve793: I used a diff tool to discover this. The request is mostly for the git history. Option 1 would be nice.

python/segyio/field.py
yngve793: Fine

lib/src/segy.c
yngve793: So, it is basically the definition of a bad variable name. :-)

achaikou: Well, in our case it is an alternative to `buffer`.
I will change it if you want to!

I guess I just made my peace with it :smile: Knowing that in some languages this is a [known convention](https://stackoverflow.com/questions/6267735/what-is-the-history-of-the-variable-names-x-and-xs) also helps :smile:

yngve793: No, need. If it is used through out the repo we could keep it.

lib/src/segy.c
ajaust: Yeah, what I wrote is confusing.

The implementation of `field_data_to_scalar` looks almost like this `switch` statement here. `field_data_to_scalar` is only missing the `SEGY_ENTRY_TYPE_TIME2 ` case. I was wondering if it would make sense to add the `SEGY_ENTRY_TYPE_TIME2` case to `field_data_to_scalar` such that we could call `field_data_to_scalar`.

python/segyio/field.py
ajaust: Yes, let's revisit later if it still bothers us then. 👍

python/test/segyio_c.py
ajaust: Looks great!

lib/src/segy.c
ajaust: Ahh, thanks for the explanation. 🙂

python/segyio/segyio.cpp
ajaust: Thanks, the added comment helps a lot. 🙂

lib/src/segy.c
ajaust: Thanks for the thorough explanation. 🙂

python/segyio/segyio.cpp
ajaust: Great 👍 

Regarding "objects we plan to hand over to Python" I might have confused myself. I meant these parts

```c++
segy_entry_definition def_mant = {
       SEGY_ENTRY_TYPE_SCALE6_MANT,
       requires_nonzero_value,
       mant_name_ptr.release()
};
```

where the allocated memory and the responsibility is given to someone else. The someone might be Python or someone else. I lost track.

python/test/segy.py
ajaust: Yes, I think it is clearer with the new name. 🙂

test-data/README.md
ajaust: If I understand correctly, the following holds for the file `mapping-unparsable-lsb.sgy`:

- The file uses least significat bit numbering so it should have the suffix `lsb`.
- The mapping is unparsable (it is non ASCII characters)
- The mapping is unparsable for different reasons than in the file `mapping-unparsable.sgy`.

I forgot why the file `mapping-unparsable.sgy` had an unparsable mapping. 😅 

Assuming that my understanding is correct, my so far best (not good) idea is to name the file `mapping-unparsable2-lsb.sgy` to indicate by the filename that it is unparsable for another reason than `mapping-unparsable.sgy`. However, I think `unparsable2` is stupid (where is `unparsable1`?) and imprecise.

Summarising: I don't have any good idea for another name or a better description (yet).

python/test/segyio_c.py
ajaust: You convinced me. Let's leave it the way it is.

ajaust (approved):
It looks good to me overall. I resolved the discussions where suitable and added some replies to the ones. 

There were two or three discussions where I tried to clarify what I was trying to say. I left these discussions as unresolved so it is easier to find my reply. None of the open discussions require any further changes. 

The only open discussion that is still on my mind is the naming and description of the new test file. However, I have no immediate ideas that sound better so there is no change required. If we can come up with an improvement before we merge this PR, then that would be great. If cannot improve it, that is fine as well. I think the test file has shown its usefulness so it should be added.

python/segyio/segyio.cpp
```diff
@@ -652,11 +652,14 @@ PyObject* segyopen( segyfd* self, PyObject* args, PyObject* kwargs ) {
         return NULL;
     }
 
-    char binary[ SEGY_BINARY_HEADER_SIZE ] = {};
-    int err = segy_binheader( ds, binary );
-    if( err ) return Error( err );
+    // endianness must be set before any other header values are read
+    if( endianness != SEGY_LSB && endianness != SEGY_MSB ) {
```
yngve793, line 652: Defining SEGY_UNDIFINED=-1 in directly in SEGY_ENDIANNESS is better han hardcoding it to -1 in `segyopen` without any context. 

I expect that it already existed. Consider it.

lib/src/segy.c
```diff
@@ -1846,6 +1846,84 @@ int segy_traces( segy_datasource* ds,
...
+    );
+    if( err != SEGY_OK ) return err;
+
+    err = field_data_to_scalar( &scalar_fd, &scalar_raw );
+    if( err != SEGY_OK ) return err;
+
+    float delay_scalar;
+    if( scalar_raw == 0 ) {
+        delay_scalar = 1.0f;
+    } else if( scalar_raw < 0 ) {
```
yngve793: Do we need the "else if" and "else". 
Can we use `1.0f * absolute(scalar_raw)`?

lib/src/segy.c
```diff
@@ -1117,11 +1117,34 @@ static int get_field( const char* header,
...
+            uint64_t val;
+            memcpy( &val, header + offset, vsize );
             val = be64toh( val );
-            memcpy( &(fd->value.f64), &val, sizeof( uint64_t ) );
+            memcpy( &( fd->value.f64 ), &val, vsize );
+            return SEGY_OK;
+        }
+
+        case SEGY_IEEE_FLOAT_4_BYTE: {
+            uint32_t val;
```
yngve793, line 1128: Declaring variables inside a switch has previously in this project caused segmentation fault and all kind of misery. Please make sure that this is safe and that you are comfortable doing this.

yngve793 (approved):
Added some new questions/comments.

python/segyio/segyio.cpp
achaikou: I am not sure I agree with it conceptually and I think working on this now would take relatively too much time.
- same refers to `encoding`.
- choice between SEGY_UNDEFINED and SEGY_ENDIANNESS_UNDEFINED name, which is not 100% straightforward.
- Python code must be updated too. Python user might call `segyopen(endianness=666)`  so then if we want to have here code `if endianness != SEGY_UNDEFINED` we have to assure that user-provided endianness is inside 3 required values or code would be incorrect.

So as this is a conceptual change from local invalid value to global invalid value I would prefer to leave the investigations to a better time...

lib/src/segy.c
achaikou: No.
It technically looks the same, but conceptually those are two different things.

`scalar` itself cannot be time2 type. It should be int2/uint2.
This delay value to which scalar is applied is on the opposite expected to be time2, we just added int2/uint2 to list of available options per PO's request.

lib/src/segy.c
achaikou: No, I think current code is most readable and straightforward.
"Absolute" will not work:
`scalar_raw == 0`: 1.0f * abs(0) = 0, expected 1
`scalar_raw == -10`: 1.0f * abs(-10) = 10, expected 0.1

python/segyio/segyio.cpp
achaikou: Option 1 - done!

lib/src/segy.c
achaikou: Yes, it is 99 matches.
Leaving as is.

lib/src/segy.c
achaikou: I am very comfortable with this!
I think previous problem was caused by not declaring a block
case AAA **{**
**}**

If local block was not declared I imagine weird stuff could happen. With local scope all should be fine.

test-data/README.md
achaikou: Your understanding is correct.

> I forgot why the file mapping-unparsable.sgy had an unparsable mapping. 😅

It says xml is unparsable. Meaning it probably is missing closing `>` or similar. 

So I've spent additional 1.5 hours thinking about it and decided that file is too old.
The root-root cause is long gone. Changing this file to align the failure reason with new code fails way earlier than original failure and thus is not interesting.
Naming this bug1 is also stupid.

So just let it burn. :fire:

ajaust (approved):

PR #656, opened 2025-11-17
Introduce support for Python 3.14

ajaust:
Fix RefTrace's garbage collection had to be adapted due Python 3.14's changes to object reference counting. I also enabled Python 3.14 as part of the wheels build again.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -53,7 +53,7 @@ jobs:
           CIBW_ENVIRONMENT_WINDOWS: >
             CMAKE_GENERATOR="${{ matrix.cmake_generator }}"
             CMAKE_GENERATOR_PLATFORM="${{ matrix.cmake_generator_platform }}"
-          CIBW_SKIP: cp38-* cp314* *-musllinux_*
+          CIBW_SKIP: cp38-* *-musllinux_*
           CIBW_ARCHS: ${{ matrix.arch }}
         run: |
             python -m cibuildwheel --output-dir wheelhouse python/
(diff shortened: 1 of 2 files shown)


python/segyio/trace.py
```diff
@@ -412,6 +412,11 @@ class RefTrace(Trace):
     :attr:`Trace.ref`. This feature requires a context manager, to guarantee
     modifications are written back to disk.
     """
+
+    # A truly internally held reference to a trace has count 2 (Python >=
```
achaikou: I am trying to understand that sentence for 15 minutes now and failing.

Maybe change to something like

> If there are no external references to the data (so only internal references remain), the reference count is
> - 2 (Python >= 3.14)
> - 3 (Python < 3.14)

?

I also question the placement of this const. If I understand correctly, the 2 references come from `self.refs` and `x`
While `self.refs` is a class attribute, `x` is just a local variable. Without it refcount is one less.
So maybe this constant should be code local to the `flush` method?

ajaust: I like your rephrased version. :) 

Regarding the placement of the variable. I was back and forth were to put it. I had it as class local and as static class variable.

I added a fixup at the end that moves the variable into the `flush` method so we can see how that looks and feels.

This was my rough though process:

### Local variable

#### Pro

- Defined close to where it is needed
- The relevant number for garbage collection is dependent on the implementation of `flush` (due to iterative over the `items`) 

### Static class variable

#### Pro

- Indicates relevance of the value for the class, i.e, the threshold is the same for any class instance.
- Python version is only evaluated once. I guess this is premature optimisation. 😁

ajaust (commented):
Thanks for the feedback. I updated the comment and the commit message and added one fixup to move the new variable into the `fetch` method.

PR #657, opened 2025-11-22
Update python interface with rev2 features

achaikou:
1.
Python 3.9 removed in this PR because [original run](https://github.com/achaikou/segyio/actions/runs/19599389300) failed for some reason, no clue why.

2.
benchmarks are failing on my test runs.
header retrieval is slower for some reason, but most importantly "create" is way slower.
Looks to be caused by "Update Field to support trace headers" commit.

I don't see anything stupid, but please let me know if you do. We need to dig deeper as it is probably not a random failure.

Upd:
I managed for now to get slowdown to a more reasonable percentage, but the problem still stays.
Apparently `__setattr__` (and maybe `__getattr__` , didn't check that or the other classes with `.name` support) is just slow.
Simply defining it adds 25% slowness to 'create' benchmark test.

We need to discuss and find out if this is a deal breaker for our interface as Field is a main class responsible for fields extraction.


python/segyio/trace.py
```diff
@@ -992,3 +992,85 @@ def __setitem__(self, i, val):
                 if isinstance(text, Text):
                     text = text[0]
                 self.segyfd.puttext(i, text)
+
+
+class Stanza(Sequence):
+    """Access stanza data from extended textual headers.
+
+    Reading stanzas is done with []. Keys are stanza indecies, values are raw
```
ajaust: `indecies` -> `indices`.

python/segyio/segy.py
```diff
@@ -375,6 +376,40 @@ def unstructured(self):
...
+
+        >>> layout = f.tracefield.SEG00000
+
+        List all field names in a specific layout:
+
+        >>> f.tracefield.SEG00000.names()
+
+        Find offset of a specific field:
+
+        >>> f.tracefield[0].xline.offset()
```
ajaust, line 415: Could I also say `f.tracefield.SEG00000.xline.offset()`?
If yes, should we add it as an example?

python/segyio/field.py
```diff
@@ -340,7 +340,7 @@ def __getitem__(self, key):
 
         Returns
         -------
-        value : int or dict_like
+        value : number, bytearray or dict_like
```
ajaust: What is the difference between an `int` and a `number`? Could it be any number type like `float`?

python/segyio/segy.py
```diff
@@ -427,12 +427,12 @@ def header(self):
...
     def header(self, val):
-        """headers macro assignment
+        """standard headers macro assignment
 
-        A convenient way for operating on all headers of a file is to use the
-        default full-file range.  It will write headers 0, 1, ..., n, but uses
-        the iteration specified by the right-hand side (i.e. can skip headers
-        etc).
+        A convenient way for operating on all standard headers of a file is to
+        use the default full-file range.  It will write headers 0, 1, ..., n,
```
ajaust: I am considering if we should explicitly mention `trace` in the docstring to reduce ambiguity since header `0, 1,...` could be also refer to an extended trace header.
```suggestion
        use the default full-file range.  It will write headers of trace 0, 1, ..., n,
```

python/segyio/segy.py
```diff
@@ -427,12 +427,12 @@ def header(self):
...
+        but uses the iteration specified by the right-hand side (i.e. can skip
+        headers etc).
```
ajaust: ?
```suggestion
        traces etc).
```

python/test/segy.py
```diff
@@ -367,10 +367,10 @@ def test_disabled_methods_seismic_unix(openfn, kwargs):
...
 def test_header_dict_methods(openfn, kwargs):
     with openfn(**kwargs) as f:
-        assert 90 == len(list(f.header[0].keys()))
-        assert 90 == len(list(f.header[1].values()))
-        assert 90 == len(list(f.header[2].items()))
-        assert 90 == len(list(f.header[3]))
+        assert 92 == len(list(f.header[0].keys()))
+        assert 92 == len(list(f.header[1].values()))
+        assert 92 == len(list(f.header[2].items()))
+        assert 92 == len(list(f.header[3]))
```
ajaust: These numbers increased by two because we recognize `Unassigned1` and `Unassigned2` as fields now. Is my understanding correct?

python/segyio/segy.py
```diff
@@ -462,6 +463,55 @@ def header(self, val):
...
+
+        Examples
+        --------
+        Read field values from trace header extension 1 at trace 5:
+
+        >>> traceheader = f.traceheader[5][1]
+        >>> traceheader.iline
+        ... 1005
+        >>> xline = traceheader.xline
+        ... 5
```
ajaust: Are `iline` and `xline` defined for the trace header extension 1? I could not find it in  the standard. Maybe we should change `iline` and `xline` to something that is defined in the trace header extension 1.

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
...
+    def __init__(self, segyfile, trace_index):
+        self.segyfile = segyfile
+        self.trace_index = trace_index
+        super().__init__(segyfile.traceheader_count)
+
+    def __getitem__(self, i):
+        """traceheader[i]
+
+        ith header of the trace, starting at 0.
+        Loads trace header into memory, so store it for access to multiple fields.
```
ajaust:
```
[...] , so store it for access to multiple fields.
```
I am not completely sure what this means. Do I (as in the user) need to store something?

python/test/segy.py
```diff
@@ -980,6 +980,105 @@ def test_write_header(small):
...
+        assert isinstance(f.traceheader[0][0], Field)
+
+        assert 1 == f.traceheader[0][0][189]
+        assert 5 == f.traceheader[-1][0][189]
+
+        with pytest.raises(KeyError):
+            f.traceheader[0][0][700]
+
+        with pytest.raises(KeyError):
+            f.traceheader[0][0][700] = 1
```
ajaust, line 995: Does this test anything new compared to the test in line 991f?

python/test/segy.py
```diff
@@ -980,6 +980,105 @@ def test_write_header(small):
...
+            f.traceheader[0][:][189] = 42
+
+        f.traceheader[1][0] = f.traceheader[0][0]
+        f.flush()
+
+        assert f.traceheader[1][0] == f.traceheader[0][0]
+        assert f.traceheader[1].SEG00000 == f.traceheader[0].SEG00000
+
+        f.traceheader[2][0:1] = f.traceheader[2][0:1]
+        assert list(f.traceheader[2][0:1]) == list(f.traceheader[2][0:1])
```
ajaust: What are we testing here?

python/segyio/segy.py
```diff
@@ -405,6 +405,16 @@ def tracefield(self):
...
 
+        Get values of field for many traces:
+
+        >>> f.tracefield.SEG00001.dt[0, 10, 20]
+
+        Note:
+
+        - this works similar to :meth:`.attributes`
+        - use :meth:`.traceheader` if you need to get many fields from the
+          trace.
```
ajaust: Is the distinction one field value from many traces vs. many field values from a single trace? Then I would consider rephrasing 
```suggestion
        - use :meth:`.traceheader` if you need to get many fields from a single
          trace.
```

python/segyio/field.py
```diff
@@ -28,6 +28,9 @@ class Field(MutableMapping):
 
     .. versionchanged:: 1.6
         more common dict operations (MutableMapping)
+
+    .. versionchanged:: 2.0
+       Support for revision 2.1
```
ajaust:
```suggestion
       Support for SEG-Y revision 2.1
```

python/segyio/field.py
```diff
@@ -374,13 +386,44 @@ def __getitem__(self, key):
...
+        -----
+        .. versionadded:: 2.0
+
+        Examples
+        --------
+        Read a single value:
+
+        >>> d.linetrc
+        1005
+        """
```
ajaust: Did this code disappear again? I could not find it in website generated by Sphinx.

python/segyio/field.py
```diff
@@ -431,13 +474,51 @@ def __setitem__(self, key, val):
...
+            The value set
+
+        Notes
+        -----
+        .. versionadded:: 2.0
+
+        Examples
+        --------
+        Set a value:
+        >>> header.linetrc = 1006
```
ajaust: I don't seem to find this docstring in my Sphinx docs either.

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
...
+            return gen()
+
+    def __getattr__(self, name):
+        """traceheaders.name
+
+        Header of the trace by name `name`. `name` must be a name defined by
+        the file layouts.
+
+        Wrapper around :meth:`.__getitem__`, so refer to it for more
+        information.
```
ajaust, line 939: I am thinking about whether I would like to remove the word "it" just to be super clear in this and similar phrases:
```suggestion
        Wrapper around :meth:`.__getitem__`. Refer to :meth:`.__getitem__` for more
        information.
```

ajaust (commented):
This looks very nice overall. 🥳  I left some suggestions or questions here and there. Mostly about the docstrings. I also build the Sphinx documentation locally. The generated documenation looks like expected.


### Regarding the failing test in Python 3.9

It complains about `  FAILED test/segy.py::test_field_via_traceheader - Failed: DID NOT RAISE <clas..`. I am not sure if we should be worried that some expected exception is not thrown, but I am also not sure if investigating it further makes sense.

### Building with Sphinx

I build it with Python 3.14 and ran into the following issues

- `requirements-doc.txt` does not specify the ReadTheDocs theme as dependency. We would need to add `sphinx-rtd-theme` in the dependency file or one needs to use `requirements-readthedocs.txt` where the dependency is specified. We could also consider consolidating both the requirement files into one.
- Building the docs on failed since `conf.py` uses distutils for `LooseVersion`. `distutils` was deprecated in Python 3.10 and [removed in 3.12](https://docs.python.org/3/library/distutils.html). I replaced it with [`packaging.version.Version`](https://github.com/Azure/azure-cli/issues/17170) which resolved the issue.

python/segyio/trace.py
achaikou: Thanks! I remember I misspelled that word somewhere, but couldn't find where and my spell checker does not work correctly anymore, for some reason :smile:

python/segyio/segy.py
achaikou: I think there are many variations of those already from before, but ok.

python/segyio/field.py
achaikou: Yes, that was the idea.
Hm. I was absolute sure this is a standard notation, so I didn't even check.
No idea where I took it from.

I think I should explicitly change to it `int, float,`

python/segyio/segy.py
achaikou: No, I think "headers" is correct.

We kind of talking about writing headers and this sentence probably means that some _headers_ are not written.

python/test/segy.py
achaikou: Yes.
Even though with separation of old Field and new Field, I will revert those.
Let old code stay old and ignore those fields.

python/segyio/field.py
achaikou: Oh, I forgot to add that one. Thanks!

python/segyio/segy.py
achaikou: I probably first wrote it with SEG00000.

python/segyio/trace.py
achaikou: Rephrased.

python/segyio/trace.py
achaikou: I think I prefer version with `it`.
Double-referencing irritates me here and I can't come up with a simple sentence instead.

Fight me on it if you prefer something else :slightly_smiling_face:

python/test/segy.py
achaikou: 991 tests `getitem` function.
995 tests `setitem` function.

python/test/segy.py
achaikou: Supposedly that we can assign slices on the traceheader level.
But I see that I forgot to change the trace number.
Thanks!

achaikou (commented):
Thanks for the blazing fast review! :fire: 

Python 3.9: I don't see a point in investigating it unless someone thinks it is useful.

Sphinx:
- last commit that touched `requirements-doc.txt` introduced it on purpose. I am not too deep into it to say if it was warranted or not, so I am inclined to leave it as is.
- thanks for the investigation! Applied.

----
### Performance

So running benchmark create test with `f.traceheader[tr][0].update({d})` is roughly 50% slower than `f.header[tr].update({d})`.

Reading this

    with segyio.open(filepath) as f:
        for tr in f.ilines * f.xlines:
            f.traceheader[tr][0][189]

vs
            
    with segyio.open(filepath) as f:
        for tr in f.ilines * f.xlines:
            f.header[tr][189]
            
is around 2 times slower.

From what I see slowness happens simply from default definition of `__setattr__`  (probably `__getattr__` too) for `TraceHeaderSequence` and `TraceHeaderField` interfaces, though I can't find any similar complaints online. I am running with 3.13 and all the issues seem to relate to 3.11
Maybe those are just fluctuations? Doubt that, given that it affected `header` test that is index-based, but still...

I am still not convinced we make the right decision to leave it be, because if slowness is real the only reasonable way to improve the performance is to remove the functions altogether, which would be a major breaking change for the users.

It probably must hold though, given that we don't have much time, but we need to highlight that in pre-release notes.

python/segyio/field.py
ajaust: I also think I have see `number` and `number-like` so from my side you are free to decide what the best type is to describe the return type.

python/segyio/trace.py
ajaust: 😨 No fight needed. Let's stick to the current version.

python/test/segy.py
ajaust: Thanks. :) 
I was not sure if Python would call `getitem` before calling `setitem` when doing and assignment.

ajaust (approved):
Great! Thanks for the fixes. It looks good to me. I just left one or two comments as "unresolved" such that my reply is easier to find. I consider everything as resolved. I also really like that you updated the Sphinx build process. In this sense: "Approved" 🎉 

I understand that the performance issue is a potential problem. Do you have some pointer to the issues that were reported in Python 3.11? Should the performance issue also show up in our benchmark tests? At least the most recent benchmark run passed successfully.

python/segyio/field.py
```diff
@@ -183,12 +183,10 @@ class Field(MutableMapping):
         'unas2'    : BinField.Unassigned2,
     }
 
-    def __init__(self, segyfile, buf, kind, traceno=None, traceheader_index=None):
-        # segyfile and traceheader_index must be initialized before everything
-        # else because setattr is overridden
-        self.segyfile = segyfile
+    def __init__(self, segyfile, buf, kind, keys, traceno=None, traceheader_index=None):
```
yngve793, line 186: It is probably well known: 
commit 1facd77d1ab855575d870a32d1b37992c554593b
Breaks test 
	  8 - python.unit (Failed)

python/segyio/field.py
```diff
@@ -374,13 +386,44 @@ def __getitem__(self, key):
         { 37: 5, 189: 2484 }
         """
 
-        traceheader_index = 0
+        traceheader_index = self.traceheader_index
```
yngve793: Value for `traceheader_index` seams to be unchanged. 
What about dropping the definition and only using `self.traceheader_index` directly?

python/segyio/field.py
```diff
@@ -520,7 +601,7 @@ def update(self, *args, **kwargs):
 
         buf = bytearray(self.buf)
 
-        traceheader_index = 0
+        traceheader_index = self.traceheader_index
```
yngve793: Can we just use` self.traceheader_index`?

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
...
+                    "Unassignable type. f.traceheader[i] can only be assigned f.traceheader[j]")
+            self[trace_index][:] = val
+            return
+
+        trace_indices = range(*i.indices(len(self)))
+        for trace_index, value in zip(trace_indices, val):
+            self[trace_index] = value
+
+
+class TraceHeaderSequence(Sequence):
```
yngve793: Following the existing naming schemes it should not be necessary to specify that it is a Sequence.

What about `TraceHeaderDefintions` or `TraceHeaderSpesifications` ?

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
...
+
+        Read a field in the first 5 headers:
+
+        >>> [x[25] for x in traceheader[:5]]
+        [1, 2, 3, 4]
+        """
+        try:
+            traceheader_index = self.wrapindex(i)
+            return Field.trace(
+                traceno=self.trace_index, traceheader_index=traceheader_index, segyfile=self.segyfile
```
yngve793: Use `self.wrapindex(i)` directly.

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
...
+        Writing fields via `traceheader[trace_index][traceheader_index] =
+        {(offset: value)}` is not supported. Use :class:`segyio.field.Field`
+        interface instead:
+
+        >>> f.traceheader[0][0].update({ 37: 5, 1: 2484 })
+
+        """
+        if not isinstance(i, slice):
+            traceheader_index = self.wrapindex(i)
+            field_sequence = self[traceheader_index]
```
yngve793, line 984: `field_sequence = self[self.wrapindex(i)]`?

python/segyio/trace.py
```diff
@@ -773,6 +773,252 @@ def xline(self, value):
         for i, src in zip(self.segyfile.xlines, value):
             self.xline[i] = src
 
+
+class TracesTraceHeaderSequence(Sequence):
```
yngve793: Not the best name. I have no real good alternative.  :-(
It seams like we are missing a term for the trace headers defined for a specific trace. 
If we had a name or word for this I think it would be easier to name this class.

python/segyio/trace.py
```diff
@@ -992,3 +1246,380 @@ def __setitem__(self, i, val):
...
+        int
+            Layout index.
+        """
+        return self.traceheader_index
+
+    def __repr__(self):
+        return f"TraceFieldLayoutSequence({self.names()})"
+
+
+class TraceFieldEntry():
```
yngve793: Very clean and nice class. :-)

python/segyio/trace.py
```diff
@@ -992,3 +1246,380 @@ def __setitem__(self, i, val):
...
+        """
+        return self._layout_names
+
+    def __repr__(self):
+        return f"TraceHeaderLayoutSequence({self._layout_names})"
+
+
+class TraceFieldLayoutSequence(Sequence):
+    """
+    Sequence of fields in a trace header.
```
yngve793: Sequence of TraceFieldEntry in a trace header?

python/segyio/trace.py
```diff
@@ -992,3 +1246,380 @@ def __setitem__(self, i, val):
...
+        list of str
+            Names of all traceheaders.
+        """
+        return self._layout_names
+
+    def __repr__(self):
+        return f"TraceHeaderLayoutSequence({self._layout_names})"
+
+
+class TraceFieldLayoutSequence(Sequence):
```
yngve793: `TraceFieldEntrySequence` or just `TraceFieldLayout`

My preference is `TraceFieldLayout`

yngve793 (approved):
In general the PR is fine. I have no strong objections. Some minor comments and suggestions are added. The only point I want to stress is the naming of classes in the  python/segyio/trace.py file. The current naming schemes works reasonably fine, but provides little conceptual understanding to the reader. What about something along these lines:

- TraceFieldEntry : Name, offset, type and value for a field
- TraceFieldSequence, TraceFieldList, TraceFieldDefinition, etc: Trace header definition as Standard, Extended etc.
- TraceFieldLayout : A sequence of TraceFieldDefinition containing Standard, Extended and private for a specific trace. 
- TraceFieldOverview: Sequence of TraceFieldLayouts for all traces. 

At least consider something along this lines.

python/segyio/field.py
achaikou: I changed it to ints and floats.

python/segyio/field.py
achaikou: As it is a fixup, no I had no clue.
But it is related to fixup-ing, should be fine with rebase.

python/segyio/trace.py
achaikou: As discussed, this became `HeaderLayoutEntries`

python/segyio/trace.py
achaikou: It is "field layout entries" now.

I didn't want to mention class names there much, as users are supposed to not pay any attention to them and not know them.

python/segyio/trace.py
achaikou: In order to have some consistency between classes we ended up with `RowFieldValues`

It is horrible, but everything else that we were discussing was at least as horrible in its own way :smile:

python/segyio/trace.py
achaikou: Was ok in other place, but here I protest.
separate variable here gives me the context - I know what `self.wrapindex(i)` is supposed to mean now.

achaikou (commented):
Thanks! :heart_eyes_cat: 

Name changes and other stuff is done directly and not in fixups, as we agreed.
Hopefully I updated everything correctly :smile:

python/segyio/trace.py
```diff
@@ -1596,3 +1598,27 @@ def __str__(self):
...
+        information.
+
+        Parameters
+        ----------
+        key : int or slice
+            Index or slice.
+
+        Returns
+        -------
+        array of numbers//bytearray
```
ajaust: Is this type `numbers//bytearray` with `//` on purpose? I am not sure what `numbers//bytearray` is supposed to mean.

python/segyio/segy.py
```diff
@@ -428,12 +428,12 @@ def header(self):
...
     def header(self, val):
-        """headers macro assignment
+        """standard headers macro assignment
 
-        A convenient way for operating on all headers of a file is to use the
-        default full-file range.  It will write headers 0, 1, ..., n, but uses
-        the iteration specified by the right-hand side (i.e. can skip headers
-        etc).
+        A convenient way for operating on all standard headers of a file is to
+        use the default full-file range.  It will write headers of trace 0, 1,
```
ajaust: Not that important. It probably renders correctly, but there is a double space:
```suggestion
        use the default full-file range. It will write headers of trace 0, 1,
```

python/segyio/field.py
```diff
@@ -28,6 +28,9 @@ class Field(MutableMapping):
 
     .. versionchanged:: 1.6
         more common dict operations (MutableMapping)
+
+    .. versionchanged:: 2.0
+       Support for SEG-Y revision 2.1
```
ajaust: Just a temporary marker to remind me where I interrupted the review.

ajaust (approved):
Looks good to me. I only found 2 minor things. :grinning: 

I did not check the documentation website generate by sphinx after the rebase/squash yet. Just give me a sign if you want me to do that. :mag:

python/segyio/trace.py
```diff
@@ -992,3 +992,85 @@ def __setitem__(self, i, val):
...
+        """Get all stanza names. Names are the same as they appear in the file.
+
+        Returns
+        -------
+        names : list of str
+            List of all stanza names in the file in order of appearance
+        """
+        return self._names
+
+    def __getitem__(self, key):
```
ajaust, line 1290: I am sorry for this additional comment, but I am checking the sphinx documenation on realised the following:

Sphinx created the documentation in the following order:
- Internal methods (`def __name__`) ordered alphabetically.
- Public methods (`def name`) ordered alphabetically.
 
The source code has a different ordering at the moment. How do you feel about ordering the methods in the new classes in the same order?

python/segyio/field.py
```diff
@@ -552,3 +552,105 @@ def trace(cls, traceno, traceheader_index, segyfile):
...
+            The value set
+
+        Notes
+        -----
+        .. versionadded:: 2.0
+
+        Examples
+        --------
+        Set a value:
+        >>> header.linetrc = 1006
```
ajaust, line 631: This does not render as expected. It does not generate a code block for the Python REPL command. Does this need an extra newline?

```suggestion
        Set a value:
        
        >>> header.linetrc = 1006
```

python/segyio/trace.py
```diff
@@ -773,6 +773,256 @@ def xline(self, value):
...
+        ----------
+        i : int or slice
+
+        Returns
+        -------
+        traceheaders : RowFieldAccessor
+
+        Examples
+        --------
+        Accessing headers in trace 10:
```
ajaust: Maybe we should repeat again (in other words) to emphasize that counting starts at 0. Something like this maybe?
```suggestion
        Accessing headers in trace 10 (the 11th trace in the file):
```

python/segyio/trace.py
```diff
@@ -773,6 +773,256 @@ def xline(self, value):
...
+        Examples
+        --------
+        Reading a header:
+
+        >>> traceheader[10]
+
+        Read a field in the first 5 headers:
+
+        >>> [x[25] for x in traceheader[:5]]
+        [1, 2, 3, 4]
```
ajaust: Why does this list have only 4 entries if it is accessing the first 5 headers?

ajaust (approved):
I just verified that the docstrings appear in the Sphinx documentation. I found some minor things (I am sorry) and added them to the review.

python/segyio/trace.py
achaikou: I think we should leave it as is - after all old classes have the same ordering.

Probably if we want to do something about it, we should do that for the whole documentation.
But I don't see any problem or need in that.

Let me know if you feel different!

python/segyio/field.py
achaikou: Yes, thank you!

I fixed so many of them, so something just had to slip from me :smile:

python/segyio/trace.py
achaikou: Good question :smile: 

Apparently I copied that from the `Header` documentation...

python/segyio/trace.py
achaikou: No, this is me hating this amount of documentation and making so many typos that I don't see them anymore :smile:

achaikou (commented):
Thanks for finding new typos! :tada:

python/segyio/trace.py
yngve793: Ok, fine.

python/segyio/trace.py
yngve793: We discussed this in person.

yngve793 (approved):
Fixups looks fine to me. 
No further comments.

python/segyio/trace.py
ajaust: No need to change it if this is consistent with the other code.

This came up because I sometimes ended up comparing docstrings with Sphinx documentation that were not for the same method, i.e, compared docstring of first method with documentation of first method (Sphinx).

ajaust (approved):
Looks good. 🎉 
Go ahead and apply the fixups and merge.

PR #658, opened 2025-11-25
Trace header extension 1

achaikou:
The only values that we would use from trace header extension 1 internally for now are `cdp_x` and `cdp_y`.
SEG00001 also overrides `offset` filed (byte 37), which we use heavily internally.
But it is overridden to a double value, when in our code it is supposed to be int.

According to PO some other field should be responsible for what we consider to be `offset`. This whole situation requires more investigations and thus is left for better days.

---

Only commits after "drop support for python 3.9" are for review, other would be removed as soon as we merge previous PR.


lib/include/segyio/segy.h
```diff
@@ -507,13 +536,18 @@ int segy_encoding( segy_datasource*, int* encoding );
...
+/* Default trace header layout map. Indices (offsets) are 0-based. */
 const segy_entry_definition* segy_traceheader_default_map( void );
-/* Default trace header name to 1-based offset map. ndicies (names) correspond
+/* Default trace header extension 1 layout map. Indices (offsets) are 0-based. */
+const segy_entry_definition* segy_ext1_traceheader_default_map( void );
+/* Default trace header name to 1-based offset map. Indices (names) correspond
  * to SEGY_FIELD enum. */
 const uint8_t* segy_traceheader_default_name_map( void );
+/* Default trace header extension 1 name to 1-based offset map. Indices (names)
+ * correspond to SEGY_FIELD enum. */
```
ajaust: Does this mention the wrong enum?
```suggestion
 * correspond to SEGY_EXTENSION1_FIELD enum. */
```

python/segyio/segyio.cpp
```diff
@@ -2466,6 +2466,38 @@ static const NameMapEntry standard_name_map[] = {
...
+    { "rec_x",     SEGY_EXT1_GROUP_X             },
+    { "rec_y",     SEGY_EXT1_GROUP_Y             },
+    { "offset",    SEGY_EXT1_OFFSET              },
+    { "nsamps",    SEGY_EXT1_SAMPLE_COUNT        },
+    { "nanosecs",  SEGY_EXT1_NANOSEC_OF_SEC      },
+    { "dt",        SEGY_EXT1_SAMPLE_INTER        },
+    { "cable_num", SEGY_EXT1_RECORDING_DEVICE_NR },
+    { "last_trc",  SEGY_EXT1_LAST_TRACE_FLAG     },
+    { "cdp_x",     SEGY_EXT1_CDP_X               },
+    { "cdp_y",     SEGY_EXT1_CDP_Y               },
```
ajaust, line 2545: Note to self: I did not double check the names. I tried to download the SEG-Y rev. 2.1 standard, but all links appear to be dead. :cry:

python/segyio/segyio.cpp
```diff
@@ -2583,23 +2631,37 @@ int initialize_traceheader_mappings(
...
+    if( extension1_index != -1 ) {
+        if( extension1_index != 1 ) {
+            auto extension1_mapping = self->traceheader_mappings[extension1_index];
+            self->traceheader_mappings.erase(
+                self->traceheader_mappings.begin() + extension1_index
+            );
+            self->traceheader_mappings.insert(
+                self->traceheader_mappings.begin() + 1, extension1_mapping
+            );
+        }
```
ajaust, line 2713: I am pretty sure that could be a [std::rotate](https://en.cppreference.com/w/cpp/algorithm/rotate.html) to the right on the range [1, extension1_index] using the reverse iterators. Obligatory [Sean Parent video](https://www.youtube.com/watch?v=UZmeDQL4LaE) which is a summary of [this talks](https://www.youtube.com/watch?v=W2tWOdzgXHA).

You have to see if you want to try using rotate. I am just happy that I remember `std::rotate` so had to mention it. :nerd_face: 

I think we could also iterate backwards from `extension_index1` to 1 and swap the mappings to move the trace header extension 1 to the index 1. This could avoid the temporary copy, but not sure if this is worth it (simpler, better?).

test-data/README.md
```diff
@@ -33,6 +33,7 @@ Due to changes in the scripts after the files were originally created, reproduci
 | text-embed-null.sgy            | small.sgy but with one byte in text header exchanged with 00                |                                                       |                                            |
 | text.sgy                       | File with unique textual file header.                                       |                                                       | no traces                                  |
 | tiny.sgy                       | Inline-sorted file with 4 samples, 3 inlines, 2 xlines.                     | make-file.py small.sgy 4 1 4 20 22                    |                                            |
+| trace-header-extension1.sgy    | File has trace header extension 1, but no explicit layout mapping.          |                                                       |                                            |
```
ajaust, line 36: What does it mean that there is "no explicit layout mapping"? Does it mean that the trace header extension 1 is the same as defined in the standard?

I am asking since in the test 

```python
f.traceheader[0].SEG00001.linetrc
```

is used. This means to me that some type of trace header extension mapping must exist.

python/segyio/segyio.cpp
```diff
@@ -689,33 +690,54 @@ PyObject* segyopen( segyfd* self, PyObject* args, PyObject* kwargs ) {
...
+    std::vector<segy_header_mapping> ordered_mappings;
+    for (const auto& header_name : traceheader_names) {
+        bool found = false;
+        for (const auto& mapping : self->traceheader_mappings) {
+            if (std::strncmp(mapping.name, header_name.data(), 8) == 0) {
+                ordered_mappings.push_back(mapping);
+                found = true;
+                break;
+            }
+        }
```
ajaust, line 746: If I understand it correctly, we return an error if the trace header is not defined in the XML trace header layout specification. If the the XML trace header layout specification contains trace header specifications, e.g., `UNUSED42`, it is fine.

Is my understanding correct?

python/segyio/open.py
```diff
@@ -110,6 +112,10 @@ def open(filename, mode="r", iline = None,
         into latin-1 encoding as per SEG-Y specification. For ascii encoding
         data would be returned as is.
 
+    layout_xml: bytearray
```
ajaust, line 115: Is the `xml_layout` simply the XML file read into a buffer? 

I realised that I have little memory of our XML parsing. I guess we get the raw XML as buffer and then parse it internally to create our favorite data structure to describe the layout.

python/test/segy.py
```diff
@@ -2639,4 +2639,43 @@ def test_revision_2_1_file_creation(tmpdir):
...
+        spec.ext_headers = 0
+        spec.traceheader_count = src.traceheader_count
+        spec.tracecount = src.tracecount
+
+        layout = src.stanza[0]
+
+        with segyio.open(small, "r") as f:
+            assert len(f.tracefield[0].names()) == 92
+        with segyio.open(small, "r", layout_xml=layout) as f:
+            assert len(f.tracefield[0].names()) == 8
```
ajaust, line 2668: This is probably my lack of memory or understanding of the test files we have:

What is the layout of `orig` such that the number of names is 8 when we use its layout instead of the default layout or the layout that is already in the file? I checked the `test-data/README.md`, but this did not help me either except that I know that there are three trace header definitions in the `orig` file.

python/segyio/segyio.cpp
ajaust, line 746: If I understand it correctly, we iterate through the trace headers and the mappings read from the XML and check if their names are in the same order. If not, the mappings are reordered to match the order of the trace headers.

I would consider rephrasing the commit message. I initially expected that the code would check that the order is the same and throw an error if this is not the case. This is what I interpreted as assuring. Therefore, I would suggest mentioning that the mappings may be reordered.

python/examples/make-rotated-rev2.py
```diff
@@ -0,0 +1,60 @@
...
+
+def product(f):
+    return itr.product(range(len(f.ilines)), range(len(f.xlines)))
+
+# this program turns the rev1 file into rev2 file with CDP values set to
+# different set of cdp-x and cdp-y coordinates in standard and ext1 headers.
+# values are taken from make-rotated-copies.py script
+
+def main():
+    if len(sys.argv) < 2:
```
ajaust: What would happen if there are more than two input files?

Maybe use
```suggestion
    if len(sys.argv) != 3:
```

python/examples/make-rotated-rev2.py
```diff
@@ -0,0 +1,60 @@
...
+            dst.bin[segyio.BinField.SEGYRevision] = 2
+            dst.bin[segyio.BinField.MaxAdditionalTraceHeaders] = 1
+
+            for i in range(src.tracecount):
+                dst.traceheader[i][1].header_name = bytes("SEG00001", "ascii")
+
+    with segyio.open(dstfile, 'r+') as dst:
+        for i, (x, y) in enumerate(product(src)):
+            trh = dst.traceheader[i][0]
+            # right
```
ajaust: What do the comments saying `left` and `right` mean here? Are these formulas equivalent to the `left` and `right` (rotation?) formulas?

python/examples/make-rotated-rev2.py
```diff
@@ -0,0 +1,60 @@
...
+import segyio
+import segyio.su as su
+
+
+def product(f):
+    return itr.product(range(len(f.ilines)), range(len(f.xlines)))
+
+# this program turns the rev1 file into rev2 file with CDP values set to
+# different set of cdp-x and cdp-y coordinates in standard and ext1 headers.
+# values are taken from make-rotated-copies.py script
```
ajaust: Maybe replace "Values" with formulas. At first, I expected an array with values.
```suggestion
# The formulas are taken from make-rotated-copies.py script
```

python/test/tools.py
```diff
@@ -174,6 +174,20 @@ def test_rotation_lsb():
...
+def test_rotation_ext1():
+    # note that in the test file ext1 values are taken from "left" and standard
+    # header values are taken from "right", yet the result does not correspond
+    # to "left" results. This happens because some of the values in the ext1
+    # header are 0 and 'use-if-non-zero' flag is set for ext1 header. So these
+    # values fall down to 'right' standard header instead.
+    #
+    # While in reality it is likely that standard header would be 0-ed, for us
+    # it is good opportunity to test 'use-if-non-zero' flag. It was checked that
+    # test result values correspond with the mixed left-right picked data.
```
ajaust, line 186: Do we also have some test that checks the `use-if-none-zero` feature in isolation?

test-data/README.md
```diff
@@ -57,6 +57,12 @@ To recreate:
 | left-small.sgy      | 270°                                                                  |
 | inv-acute-small.sgy | 315°                                                                  |
 
+File below is used to test revision 2 features and is created with:
+`python python/examples/make-rotated-rev2.py test-data/small.sgy test-data/rotated-small-rev2.sgy`
+
+| File                   | Purpose                                                                   |
+|------------------------|-------------------------------------------------------------------------- |
+| rotated-small-rev2.sgy | Revision 2 file with different values on standard and extension 1 levels. |
```
ajaust: Why do we introduce the new name "level" here and what does it mean? Could we use "headers" here? This table is new so we could also write more if this makes the purpose clearer.
```suggestion
| rotated-small-rev2.sgy | Revision 2 file with different values on standard and extension 1 headers. |
```

ajaust (commented):
I think overall it looks good. I only have a few smaller comments.

python/segyio/segyio.cpp
achaikou: :unamused: 
Looks like they moved [here](https://seg.org/Publications/SEG-Technical-Standards/)...

So we need another commit to add to both our release branches to change Readme values... Again.
I hope they won't change those again soon...

python/segyio/segyio.cpp
achaikou: :smile: 
I recall seeing suggestion with `rotate`, but did not understand the point and it seems to me less readable.
The backwards iteration I do not recall, but also do not understand :smile: 

Given that this is a one-time per file operation, also very likely never-happening, I would prefer erase-insert for readability. :blush: 

P.S. very enlightening `rotate` summary :smile:

test-data/README.md
achaikou: Exactly!
There is no D8 xml, but trace header extension 1 is present in the actual file headers.

So this file checks that we use the default SEG00001 mapping correctly when no mapping override is provided.

python/segyio/segyio.cpp
achaikou: Yes. :+1: 
`UNUSED42` entry is totally fine. We return an error only if header is used in the file, but its layout is not defined.
So for example if xml is present and contains only SEG00000, but in the file SEG00001 is used, I consider that to be an error.

Seemed logical.

achaikou: Your understanding is correct.

I would change on rebase to something like

> Match mappings order with actual

python/segyio/open.py
achaikou: Yes!

That should be pure xml.
So I called it `layout_xml` to emphasize that xml is the form.

python/test/segy.py
achaikou: It doesn't really matter much.
Here we just check that you can open file with different layouts (file-owned vs custom) and get different names/results. Which layout is correct to get the true data is user's problem, if they wish to override whatever is in the file.

So layout of `trace-header-extension.sgy` is just something different from default.
In this case it is minimal-ish number of declared fields (8), but at standard positions, which allows us to open the `small` file.

python/examples/make-rotated-rev2.py
achaikou: Yes. Added a few words for clarification.

python/test/tools.py
achaikou: No. I can't image what it can be.
- `rotation` is the only place where we are using those. I don't think we can (or it makes sense) to create some better rotation setup that would focus on use-if-non-zero feature.
- There is no functionality for the user to call "give_me_cdp_x" where we internally would return value from standard or extended header. Users have use-if-non-zero information via FieldLayoutEntry class, so it is up to them to create correct logic.

achaikou (commented):
Thanks for over-the-clock review! :cat:

lib/include/segyio/segy.h
```diff
@@ -157,6 +157,35 @@ typedef enum {
...
+    SEGY_EXT1_OFFSET                = 129,
+    SEGY_EXT1_SAMPLE_COUNT          = 137,
+    SEGY_EXT1_NANOSEC_OF_SEC        = 141,
+    SEGY_EXT1_SAMPLE_INTER          = 145,
+    SEGY_EXT1_RECORDING_DEVICE_NR   = 153,
+    SEGY_EXT1_ADDITIONAL_TR_HEADERS = 157,
+    SEGY_EXT1_LAST_TRACE_FLAG       = 159,
+    SEGY_EXT1_CDP_X                 = 161,
+    SEGY_EXT1_CDP_Y                 = 169,
+    SEGY_EXT1_TRACE_HEADER_NAME     = 233
```
yngve793, line 186: For standard trace header and file header "UNASSIGNED" is used to identify unassigned regions. Should we keep the pattern? 177?

lib/src/segy.c
```diff
@@ -321,6 +321,42 @@ static const segy_entry_definition traceheader_default_map[SEGY_TRACE_HEADER_SIZ
...
+    [ -1 + SEGY_EXT1_OFFSET                ] = { SEGY_ENTRY_TYPE_IEEE64,    true,  NULL },
+    [ -1 + SEGY_EXT1_SAMPLE_COUNT          ] = { SEGY_ENTRY_TYPE_UINT4,     true,  NULL },
+    [ -1 + SEGY_EXT1_NANOSEC_OF_SEC        ] = { SEGY_ENTRY_TYPE_INT4,      false, NULL },
+    [ -1 + SEGY_EXT1_SAMPLE_INTER          ] = { SEGY_ENTRY_TYPE_IEEE64,    true,  NULL },
+    [ -1 + SEGY_EXT1_RECORDING_DEVICE_NR   ] = { SEGY_ENTRY_TYPE_INT4,      false, NULL },
+    [ -1 + SEGY_EXT1_ADDITIONAL_TR_HEADERS ] = { SEGY_ENTRY_TYPE_UINT2,     false, NULL },
+    [ -1 + SEGY_EXT1_LAST_TRACE_FLAG       ] = { SEGY_ENTRY_TYPE_INT2,      false, NULL },
+    [ -1 + SEGY_EXT1_CDP_X                 ] = { SEGY_ENTRY_TYPE_IEEE64,    true,  NULL },
+    [ -1 + SEGY_EXT1_CDP_Y                 ] = { SEGY_ENTRY_TYPE_IEEE64,    true,  NULL },
+    [ -1 + SEGY_EXT1_TRACE_HEADER_NAME     ] = {SEGY_ENTRY_TYPE_STRING8,    false, NULL },
```
yngve793, line 357: Consider adding UNASSIGNED

lib/src/segy.c
```diff
@@ -476,10 +512,49 @@ static const uint8_t traceheader_default_name_map[SEGY_TRACE_HEADER_SIZE] = {
...
+    [ SEGY_EXT1_OFFSET                ] = SEGY_EXT1_OFFSET,
+    [ SEGY_EXT1_SAMPLE_COUNT          ] = SEGY_EXT1_SAMPLE_COUNT,
+    [ SEGY_EXT1_NANOSEC_OF_SEC        ] = SEGY_EXT1_NANOSEC_OF_SEC,
+    [ SEGY_EXT1_SAMPLE_INTER          ] = SEGY_EXT1_SAMPLE_INTER,
+    [ SEGY_EXT1_RECORDING_DEVICE_NR   ] = SEGY_EXT1_RECORDING_DEVICE_NR,
+    [ SEGY_EXT1_ADDITIONAL_TR_HEADERS ] = SEGY_EXT1_ADDITIONAL_TR_HEADERS,
+    [ SEGY_EXT1_LAST_TRACE_FLAG       ] = SEGY_EXT1_LAST_TRACE_FLAG,
+    [ SEGY_EXT1_CDP_X                 ] = SEGY_EXT1_CDP_X,
+    [ SEGY_EXT1_CDP_Y                 ] = SEGY_EXT1_CDP_Y,
+    [ SEGY_EXT1_TRACE_HEADER_NAME     ] = SEGY_EXT1_TRACE_HEADER_NAME,
```
yngve793, line 547: UNASSIGNED?

python/examples/make-rotated-rev2.py
```diff
@@ -0,0 +1,60 @@
...
+    with segyio.open(dstfile, 'r+') as dst:
+        for i, (x, y) in enumerate(product(src)):
+            trh = dst.traceheader[i][0]
+            # right
+            trh.cdp_x = y
+            trh.cdp_y = 100 - x
+            trh.co_scal = 1
+
+            trh = dst.traceheader[i][1]
+            # left
```
yngve793: Do I understand correctly that the value 0 is rotated right through the traces and that value 1 is rotated left?

yngve793 (approved):
A lot of the code is reasonably straight forward. I had a hard time finding anything of significance. :-(. Most natural questions where already asked by Alexander.

lib/include/segyio/segy.h
achaikou: I don't see a reason for it and I really don't want to add them.

Those previous unassigned fields caused some problems along the way because in newer specification version they become assigned. PO had to choose what to do with that (and he chose leaving unassigned fields be).

Given that mappings now can be custom, I think we should not add anything unassigned.

lib/src/segy.c
achaikou: Same - I think this is not needed.

lib/src/segy.c
achaikou: No?

python/examples/make-rotated-rev2.py
achaikou: Those are taken from make-rotated-copies script.
From what I understand these numbers correspond to cdp values rotated 45 degrees (left) and 135 degrees (right).
I didn't go into understanding what is going on there very closely.

python/segyio/segyio.cpp
ajaust: This sounds better. I would consider being even more explicit if it fits with the line length. Maybe something like this:

> Match mappings order with traceheader order

python/segyio/segyio.cpp
ajaust: Thanks for the link. 👍 

I checked the names and could not find any deviations.

ajaust (approved):
Looks good to me. I only had an additional suggestions for an alternative commit message, but that's it. 🙂

PR #659, opened 2025-12-02
Changelog 2.0 alpha

achaikou:


changelog.md
```diff
@@ -1,3 +1,19 @@
+# 2.0.0-alpha.1
+* Support for SEG-Y revision 2.1:
```
ajaust: Maybe rephrase it slightly milder since some features are not fully implemented (`offset` as float, reading effective value of fields that may be overwritten in trace header extension 1 like `cdp_x`).

Suggestions:
```suggestion
* Partial support for SEG-Y revision 2.1:
```

or 

```suggestion
* Initial support for SEG-Y revision 2.1:
```

changelog.md
```diff
@@ -1,3 +1,19 @@
...
+  * support for int64, uint64 and floating point types in trace headers.
+  * support for trace header extension 1 and proprietary trace headers via
+  `f.traceheader`.
+  * support for XML defined trace header layouts (see Sec. D8 in SEG-Y standard)
```
ajaust:
```suggestion
  * support for XML-defined trace header layouts (see Sec. D8 in SEG-Y standard)
```

changelog.md
```diff
@@ -1,3 +1,19 @@
+# 2.0.0-alpha.1
```
ajaust, line 1: I would consider using a more generic commit message, e.g.,

```
Bump version to 2.0.0-alpha.1
```

since we do a little bit more than only bumping the version tag.

ajaust (approved):

PR #660, opened 2025-12-02
Bump version to 1.9.14

achaikou:


changelog.md
```diff
@@ -1,3 +1,7 @@
+# 1.9.14
+* Distribute wheels for Python 3.14
+* Support for python 3.9 has been dropped, as it is end-of-life
```
ajaust: Since "Python" is capitalised in the first bullet point we could also capitalise it here. However, not capitalising it would be consistent with the previous release notes.

```suggestion
* Support for Python 3.9 has been dropped, as it is end-of-life
```

ajaust (approved):
One minor thing. You have to see if you care about changing it. 🙂

PR #661, opened 2025-12-03
After release updates

achaikou:


ajaust (approved):

ajaust (approved):
Bump of benchmark reference also looks good. :)

